PR #83, opened 2014-11-10
Changes entry Id's prior to saving a cloned patient

dczulada:


okeefm:
Change pulled in in #135

PR #123, opened 2015-01-12
Updated cypress version numbers in Packer build script

okeefm:
Bumped version numbers in Packer build script from 2.5.0 to 2.6.0. Also added variables to name the VMs, so they don't get the default Packer name.


eedrummer:
Rebased and merged

PR #124, opened 2015-01-12
bump version to 2.6

ssayer:
Bump Cypress version to 2.6. Also changes sort order in test wizard so most recent bundle is the default.


eedrummer:
Rebased and merged

PR #125, opened 2015-01-12
Cypress install script updates

okeefm:
Updated to the 2.6.0 version of Cypress as the default installed version.

Also updated some of the text output to the terminal, to format it better/hide text that shouldn't be displayed.


eedrummer:
Rebased and merged

PR #129, opened 2015-01-13
Fixes intermittent test suite failures, mainly on Travis CI

eedrummer:
The test suite will intermittently fail on Travis, but this can be
replicated locally. The ProductTestHelperTest depends on the measures
collection being populated. Depending on test execution order or
previous state of a developer's database, the collection may or may
not be populated.

To ensure this test fails, drop the cypress_test database and run:
bundle exec rake test TEST=test/unit/helpers/product_test_helper_test.rb

Since Travis always starts with a clean database, it is much more
likely to see the test fail there.

This change ensures that the measure collection will be populated
before the test is run.


okeefm:
Looks good to me.

PR #131, opened 2015-01-13
UI issues with tabs caused by jQuery

ssayer:
* Missing background images caused by incorrect paths
* Tab text overflowing instead of wrapping


okeefm:
Looks good to me.

okeefm:
Merged.

PR #132, opened 2015-01-14
Updated patients controller so @measures_categories is generated for tests

okeefm:
Before, if there was a fragment matching `index-@bundle.version` (the MPL) we wouldn't generate it, but we don't store fragments for product tests, so the @measures_categories was needed to generate the view.


eedrummer:
Rebased and merged

PR #133, opened 2015-01-14
Added dependent: :destroy to test_execution artifact

okeefm:
That way it gets deleted when the test execution does (also deletes the file(s) on disk). Keeps the data/uploads/test_executions folder from growing to consume ridiculous amounts of disk space.

### DIFF
--- a/app/models/product_test.rb
+++ b/app/models/product_test.rb
@@ -4,7 +4,7 @@ class ProductTest
 
   belongs_to :product
   has_one :patient_population
-  has_many :test_executions, dependent: :delete
+  has_many :test_executions, dependent: :destroy
   belongs_to :user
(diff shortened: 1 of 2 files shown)


ssayer:
Does this actually fire since product test used dependent :delete?

okeefm:
It seemed to fire (I've been able to delete the measure eval vendor, and it deletes all the relevant test executions) but I can change the :`delete` to `:destroy` in `product_test` if you like.

ssayer:
Yeah we should probably make that a destroy, I don't think there's a good reason to skip any callbacks.

okeefm:
Done.

eedrummer:
Rebased and merged

PR #146, opened 2015-02-05
Removed app secret token from secret_token.rb 

okeefm:
(the Rails 3 and below way of doing it) and put it in secrets.yml instead.


ssayer:
Rebased and merged

PR #148, opened 2015-02-13
Updated Rake task and install script to install over SSL

okeefm:
Because we added an SSL certificate to the server, the "right" way to connect to it is now https://

Also added changes to the keyserver definitions for RVM and Mongodb: added a backup keyserver for RVM, and added the missing `hkp://` and `:80` protocol and port definitions for the mongo keyserver. 

Includes changes from #152


ssayer:
rebased and merged

PR #149, opened 2015-02-18
Results download fix

okeefm:
Fixes the "download results" error, and the test that should have caught it.


ssayer:
rebased and merged

PR #152, opened 2015-02-18
Update install_cypress.sh

jawz101:
the keyserver was failing on me and ubuntu's keyserver worked.  Also, specifying port 80 helped if I'm being blocked on the default port 11371

### DIFF
--- a/contrib/install_cypress.sh
+++ b/contrib/install_cypress.sh
@@ -490,7 +490,7 @@ done
 
 # Install the RVM GPG keys
 echo -n "   Install RVM GPG Keys: "
-gpg --keyserver hkp://keys.gnupg.net --recv-keys D39DC0E3 &> /dev/null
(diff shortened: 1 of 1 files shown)


okeefm:
Hey,

Thanks for the pull request! I'm not sure why that keyserver is failing, but I'm seeing it too. My suggestion would be to set the Ubuntu keyserver as a backup to `keys.gnupg.net`, rather than just overwriting it, but I think this is a good idea.

jawz101:
I was thinking that as well.  I'm not much of a programmer.  It takes a long time to timeout if it's going to fail which is kinda annoying.
another one that works well is pgp.mit.edu and, regardless, there was a legit but on the mongodb one as it was missing the hkp://  ... maybe it isn't needed?  I dunno.  I was looking at the mongodb instructions as I was sort've manually getting past those 2 errors when I was trying this out.

ssayer:
Changes merged along with #148

PR #153, opened 2015-02-19
Refactored table_measure in the patients controller

okeefm:
Made it less complex, and DRYed it up a little.


ssayer:
rebased and merged

PR #155, opened 2015-03-03
Disabled download patients button when no patients in test

okeefm:
If there are no patients in the test, don't let them hover/click on the link to download, and change the text. Doesn't solve the issues with the measure evaluator not liking the lack of zip file, but keeps users from seeing an error page.

Also edited _markup so the xml all colors correctly.


ssayer:
Rebased and merged

PR #174, opened 2015-06-22
Submission issues

dczulada:
Created Measure Period Validator
Updated Expected Results Validator to use shared methods in HDS
Updated Expected Results Validator to use Validator module
Product Tests (Cat I's) now store the "Parent Cat 3" Ids for data criteria checking
Fixtures updated


ssayer:
I think you may need to bump the version of HDS you're referencing...

ssayer:
I rebased and squashed this, but the code is now on master:

https://github.com/projectcypress/cypress/commit/b1672d41432703a921e767567a6953caa9e5329d

PR #178, opened 2015-07-09
Test name updated to use CMS id instead of NQF id

dczulada:


ssayer:
Going to hold off this one, since it'll cause issues with patched systems.

ssayer:
Since we're moving to a new branch for Cypress v3 it doesn't make sense to make this update for v2 anymore

PR #180, opened 2015-07-10
Changed 'stratification' to 'STRAT' in expected results validator

okeefm:
We had to make this change in HDS, to accommodate QRDA Cat 3s with 'STRAT' as the stratification value, rather than 'stratification'. 

May cause issues for Cypress 2.6.1; not sure.


dczulada:
What about something like this for backward compatibility

      stratification = nil
      if _ids["STRAT"] != nil
        stratification = _ids.delete("STRAT")
      elsif _ids["stratification"] != nil
        stratification = _ids.delete("stratification")
      end

ssayer:
Yes, we need backwards compatibility because the code needs to support both 2014 and 2015 bundles

PR #200, opened 2015-10-14
Fix vendor alert and tests

jaebird123:
Mongoid 5.0 downgraded to Mongoid 4.0.2. Created all new unit tests for Vendor model. Fixed issue with vendor validation alert occurring on vendor index page.


app/views/vendors/index.html.erb
```diff
@@ -1,6 +1,11 @@
 <div class="vendors index_vendor_page col-sm-12">
 
-  <%= render :partial => 'vendor_alert' %>
+  <% if flash[:notice] %>
+    <div class = "alert alert-<%= flash[:notice_type] || 'info' %>" >
+      <%= flash[:notice] %>
+      <a href = "#" class = "close" data-dismiss = "alert" aria-label = "close">&times;</a>
+    </div>
+  <% end %>
```
thecristen: Could the validation alert partial be written more generally so that it could be used here too?

jaebird123: The vendor validation error no longer uses flash (sam made this change and it's a lot cleaner) but the notice on the index page does. I think they're pretty separate cases but it might be useful to move them to general partials for all views to use since they will probably be used in products also.

PR #208, opened 2015-10-19
Manage products

jaebird123:
Added ability to manage products. Users can create, edit, and remove products associated with a single vendor. Each product requires a unique name within its parent vendor, an EHR type, and certification test types. EHR type and certification test types cannot be edited once product is created.


app/assets/javascripts/application.js
```diff
@@ -3,7 +3,7 @@
 //= require jquery_ujs
 //= require jquery-ui
 //= require bootstrap-sprockets
-//= require bootstrap
+//////// require bootstrap
```
ssayer: Does the extra /'s make this into a commit, or does it still include bootstrap? What is the desired behavior?

app/models/product.rb
```diff
@@ -10,7 +11,26 @@ class Product
   field :version, type: String
   field :description, type: String
 
-  validates :name, presence: true
+  validates :name, presence: true, uniqueness: { scope: :vendor, message: "Product name was already taken. Please choose another." }
+  validates :ehr_type, presence: true
```
ssayer: Is this for EH/EP? If so, we should be validating those values, not just the presence

app/models/product.rb
```diff
@@ -10,7 +11,26 @@ class Product
...
+  validate :at_least_one_test_type?
   validates :vendor, presence: true
-  # methods must be added
+
+  field :ehr_type, type: String
+  field :c1_test, type: Boolean
```
ssayer: What is the purpose of these booleans? It seems repetitive if we're storing records for the measure tests...

app/assets/stylesheets/cypress/_products.scss
```diff
@@ -0,0 +1,53 @@
...
+.form-group:last-child {
+  margin-top: 1em;
+}
+
+.col-sm-1 > button, .col-sm-2 > button,
+.col-sm-3 > button, .col-sm-4 > button,
+.col-sm-1 > label, .col-sm-2 > label,
+.col-sm-3 > label, .col-sm-4 > label {
+  width: 100%;
+}
```
thecristen: It might be better to just adjust the markup to include the `.btn-block` class??

app/views/products/_product_form.html.erb
```diff
@@ -0,0 +1,109 @@
...
+  <div class = "form-group">
+    <%= f.label :description, "Description", :class => "control-label col-sm-2" %>
+    <div class = "col-sm-4">
+      <%= f.text_area :description, :class => "form-control", :placeholder => "Write product description here..."%>
+    </div>
+  </div>
+
+  <div class = "form-group">
+    <%= f.label :ehr_type, "EHR Type", :class => "control-label col-sm-2" %>
+    <div class = "col-sm-2">
```
thecristen: Both `col-sm-2` can be consolidated to a `col-sm-4` and then you can wrap the two buttons in `<div class = "btn-group btn-group-justified" role="group"></div>` and have the desired effect without extra CSS

app/views/products/_product_form.html.erb
```diff
@@ -0,0 +1,109 @@
...
+      <%= f.label :ehr_type_provider, "Eligible Provider", class: (f.object.persisted? ? "cannot_select btn btn-default" : "btn btn-default") %>
+    </div>
+    <div class = "col-sm-2" >* required</div>
+  </div>
+
+  <div class = "row">
+    <label class = "control-label col-sm-2">Certification Types</label>
+    <div class = "col-sm-4">
+      <%= f.check_box :c1_test, disabled: f.object.persisted?, :class => "hide" %>
+      <%= f.label :c1_test, "C1 Test - Record and Export", class: (f.object.persisted? ? "cannot_select btn btn-default" : "btn btn-default") %>
```
thecristen: You can add `btn-block` to these so you don't need CSS to make it full-width

app/views/products/_product_form.html.erb
```diff
@@ -0,0 +1,109 @@
...
+    <div class = "col-sm-2"></div>
+    <div class = "col-sm-4">
+      <%= f.check_box :c4_test, disabled: f.object.persisted?, :class => "hide" %>
+      <%= f.label :c4_test, "C4 Test - Data Filtering", class: (f.object.persisted? ? "cannot_select btn btn-default" : "btn btn-default") %>
+    </div>
+  </div>
+
+  <div class = "form-group">
+    <div class = "col-sm-2"></div>
+    <div class = "col-sm-1">
```
thecristen: The spacing between the buttons looks awkward and inconsistent with buttons elsewhere. I recommend consolidating the two button columns to something like 
```
<div class = "col-sm-4">
      <div class = "pull-right">
              (buttons)
     </div>
</div>
```

dczulada: If you leave the EHR type blank when creating a new product you get this message "Ehr type can't be blank."  After this message is displayed, you cannot highlight an EHR type.  Your allowed to select the EHR type (and it will save correctly), however it does not show up with the purple highlight.

app/assets/javascripts/application.js
jaebird123: Extra slashes makes it a comment. We need EITHER bootstap-sprockets or boortstrap but not both

app/models/product.rb
jaebird123: yes, you're right. i'll make that change

app/assets/javascripts/application.js
ssayer: It doesn't make sense as a comment, so you should just get rid of it if that's the case. @thecristen should know the desired behavior.

app/models/product.rb
jaebird123: the purpose of the booleans is to give the choice to the user who makes the product for which types they will be testing for. ex.) if only c1_test is true then all measure tests should include the ability to only test c1 for that measure

app/assets/javascripts/application.js
jaebird123: ok, I'll get rid of it

app/views/products/_product_form.html.erb
jaebird123: ahh, much better

PR #229, opened 2015-10-28
Updated step definitions and added cucumber tests to travis

schreiaj:


Gemfile.lock
```diff
@@ -454,6 +454,3 @@ DEPENDENCIES
   turbolinks
   uglifier (>= 1.3.0)
   web-console (~> 2.0)
-
-BUNDLED WITH
```
okeefm: Do we care if this `BUNDLED WITH` thing gets removed here/added in elsewhere?

features/step_definitions/user.rb
```diff
@@ -17,7 +17,7 @@
   visit '/users/sign_in'
   page.fill_in 'Email', with:  @user.email
   page.fill_in 'Password', with:  @user.password
-  page.click_button 'Log in'
+  page.click_button 'Sign in'
 end
 
 Then(/^the user should see an log in success message$/) do
```
okeefm: Technically it's "a" login message, rather than "an".

schreiaj: Agreed, but I didn't write those specs. I assumed there was a reason to write it that way. Or that it was standard.

okeefm: True.

ssayer: This was copy/pasted initially from a Cucumber/Devise tutorial... it can change

PR #230, opened 2015-10-28
Added integration testing for vendors and products pages

jaebird123:
Added integration testing for vendors and product pages. Added test execution folder in public directory to .gitignore. Fixed product successful notification. Changed Edit and Remove icons for products to buttons with text. Changed required fields in forms to not required since notifications will catch this with appropriate error message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,7 @@ contrib/packer_cache/
 # For built boxes
 contrib/output*
 contrib/*.tgz
+
+# test executions
+public/data*
+public/uploads*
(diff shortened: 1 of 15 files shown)


.travis.yml
```diff
@@ -23,3 +23,4 @@ script:
   - ./bin/brakeman -qAzw1
   - ./bin/overcommit --run
   - ./bin/rake test
+  - ./bin/cucumber
```
okeefm: The cucumber tests are already added to Travis in #229

PR #232, opened 2015-11-03
Initial record filtering for C4 testing

dehall:
Fills in the C4 task filtering and creates a new class to filter patient records. Provider filtering criteria as defined in the final rule are still awaiting clarification and have not been implemented yet

### DIFF
--- a/app/models/c4_task.rb
+++ b/app/models/c4_task.rb
@@ -4,8 +4,44 @@ class C4Task < Task
     # for each of the measures in the product_test but constrainted to the
     # filter provided for test.  Filter data should live in the options field
     # that is defined in Task
+    MeasureEvaluationJob.perform_now(self, 'filters' => patient_cache_filter)
   end
 
   def execute(_params)
+  end
+
+  # Final Rule defines 9 different criteria that can be filtered:
+  #
+  # (A) TIN .................... (F) Age
+  # (B) NPI .................... (G) Sex
+  # (C) Provider Type .......... (H) Race + Ethnicity
+  # (D) Practice Site Address .. (I) Problem
+  # (E) Patient Insurance
+  #
+  def patient_cache_filter
+    input_filters = (options['filters'] || {}).dup
+    filters = {}
+
+    # QME can handle races, ethnicities, genders, (providers, languages) and patient_ids
+    # so pass these through directly
+    # is it really worth it to do this? does it make more sense to just do everything based on patient id
+    # and simplify the logic here
+
+    filters['races'] = input_filters.delete 'races' if input_filters['races']
+    filters['ethnicities'] = input_filters.delete 'ethnicities' if input_filters['ethnicities']
+    filters['genders'] = input_filters.delete 'genders' if input_filters['genders']
+
+    # for the rest, manually filter to get the record IDs and pass those in
+    if input_filters.count > 0
+      filters['patients'] = Cypress::RecordFilter.filter(product_test.records, input_filters, effective_date: product_test.effective_date).pluck(:_id)
+    end
+
+    filters
+  end
+
+  # this will fetch the records associted with the product test but constrained
+  # according to the filters configured for the task
+  def records
+    Cypress::RecordFilter.filter(product_test.records, options['filters'], effective_date: product_test.effective_date)
   end
 end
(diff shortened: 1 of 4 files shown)


lib/cypress/record_filter.rb
```diff
@@ -0,0 +1,94 @@
...
+    def self.create_age_query(query, age_filter, options)
+      # filter only by a single age range, can be age < max, age > min, or min < age < max
+      effective_date = options[:effective_date]
+
+      if age_filter['max']
+        age_max = age_filter['max']
+        req_birthdate = Time.at(effective_date).utc - age_max.years
+
+        query[:birthdate.gte] = req_birthdate
+      end
```
rdingwell: If following the measures conventions of what age at measure period means this needs to account for some corner case issues.   For instance , if the filter is 18 year at Measure Period then due to the way date time logic is performed in the measures someone who is born on 01/01/xxxx 18 years from the MP still meets the measure logic even though the absolute value for the birth time compared to the start of the measure period would state otherwise.

lib/cypress/record_filter.rb
```diff
@@ -0,0 +1,94 @@
...
+
+      if input_filters['age']
+        create_age_query(query, input_filters['age'], options)
+      end
+
+      if input_filters['problems']
+        create_problem_query(query, input_filters['problems'])
+      end
+
```
rdingwell: You may want to reconsider query being a simple hash.  You can make this an $and query instead and add the other queries into an array.  So this would be something along the lines of this

queries = []
query = {'$and' => queries}

after this you can put all of the other queries into the array and also alleviate your worry declared below about possibly needing more than one $or query.  You would also simplify some of the other method calls by not needing to pass the query hash around.

lib/cypress/record_filter.rb
```diff
@@ -0,0 +1,94 @@
...
+      relevant_codes.uniq!
+
+      problem_subquery = { '$elemMatch' => { 'codes.SNOMED-CT' => { '$in' => relevant_codes } } }
+
+      conditions = { 'conditions' => problem_subquery }
+      procedures = { 'procedures' => problem_subquery }
+      encounters = { 'encounters' => problem_subquery }
+
+      # TODO: this can be dangerous, what if something else needs an OR
+      query['$or'] = [conditions, procedures, encounters]
```
rdingwell: See comment above

app/models/c4_task.rb
```diff
@@ -4,8 +4,44 @@ def after_create
...
+  # Final Rule defines 9 different criteria that can be filtered:
+  #
+  # (A) TIN .................... (F) Age
+  # (B) NPI .................... (G) Sex
+  # (C) Provider Type .......... (H) Race + Ethnicity
+  # (D) Practice Site Address .. (I) Problem
+  # (E) Patient Insurance
+  #
+  def patient_cache_filter
+    input_filters = (options['filters'] || {}).dup
```
rdingwell: where is options defined? I don't see it declared anywhere

dehall: options is inherited from the Task class

rdingwell: So it is, and I put it there.  No points for me.

rdingwell:
I'm going to merge this in though we need to add a task to address the corner case issues with the date filtering however.

ssayer:
What's the corner case issue?

dehall:
Rob gives a good example above: `if the filter is 18 year at Measure Period then due to the way date time logic is performed in the measures someone who is born on 01/01/xxxx 18 years from the MP still meets the measure logic even though the absolute value for the birth time compared to the start of the measure period would state otherwise.`
The code here is considering age as of the start of the measure period (i.e., the first second of the first day) but per the above it should really be the end of the first day. So this filter could count someone with a birthdate on the first day of the measure period incorrectly.

rdingwell:
The corner case issue arises for birthdays on 01/01/xxxx . The way that the
calculation works is that the AgeAt qdm function does not take
hours,minutes and seconds into account.  So for instance if there is a
constraint like this AgeAt measure period >= 18.     Now say there is a
patient whos birthday is  01/01/xxxx 3:00 PM and they are turning 18 on the
start of the MP.  With a straight timestamp based filtering approach this
patient would not be included because the hours, min.... would cause them
to look as if they are still 17 at the start of the MP.  However , the date
time logic calculation does not take the smaller units into account so what
will be compared is the year,month and day.  Taking that into account the
patient would be included because they will 18 at the start of the MP.

PR #234, opened 2015-11-06
Add products with tests

jaebird123:
- add product form now includes selection for task type as well as measure selection. for each measure selected, a product_test is created and attached to the product. as of now, measures cannot be added or removed after the product is created
- product show page now lists each associated measure and its status (failing, incomplete, or passing)
- status of test execution now bubbles up to parent models (child -> parent: test_execution -> task(c1, c2, c3, or c4) -> product_test -> product -> vendor). status is passing if all children are passing, failing if any one child is failing, and incomplete otherwise
- added product_test show page
- changed style of add / edit vendor form which now uses bootstrap-forms


okeefm:
Does it make any sense to squash these commits down?

ssayer:
Yes, please squash these down and i'll pull it in

PR #238, opened 2015-11-12
Add capybara accessible

thecristen:
Now your Cucumber tests will fail if anything in the pages tested fails [Google's Accessibility Developer Tools](https://github.com/GoogleChrome/accessibility-developer-tools) auditing! 

Includes a bonus commit in which I soothe some failing tests.


ssayer:
I manually merged this in via rebase to ensure this didn't break any of the previous PRs I pulled in.

PR #246, opened 2015-11-19
Added authenticate_user! before_filter to controllers, so people are …

okeefm:
Added `before_filter: authenticate_user!` to most of the controllers (not `ApplicationController`), so people are required to sign in before accessing them.


okeefm:
Never mind, Rob already fixed this.

PR #247, opened 2015-11-19
Test execution page

jaebird123:
added to measure test execution page: error reporting table, product test information, ability to download CAT 1 zip file, ability to upload CAT 1 zip file for C1 tests, separate tabs for CAT 1 vs CAT 3 executions if C1 and C2 are selected at the product level.


ssayer:
@thecristen @jaebird123  That "found two elements" thing has popped up again

PR #253, opened 2015-12-03
show error when no bundles installed

dehall:
trying this again, tests all pass locally so hopefully they work on travis


ssayer:
The bundles names and urls should be put in the config/cypress.yml file, which is where any versioned information goes.

PR #254, opened 2015-12-03
Allow users to edit measures for a product

thecristen:
Addresses https://www.pivotaltracker.com/story/show/108449030


dczulada:
Is this a true statement "Changing a product's associated measures will reset all associated tests."  When i read this, it makes it seem like it will change all tests associated with the product.

Should it be "Changing a product's measures will reset all tests associated with the modified measures."

ssayer:
Yes Dave I that is correct. We should chat about the consequences of removing a measure.

PR #268, opened 2015-12-10
Bumped rails to 4.2.5 in Gemfile

okeefm:


ssayer:
Looks like travis doesn't like bundle exec, probably need add some stuff to the before_script section of travis.ml

PR #272, opened 2015-12-14
c3 validations moved out of c1 and c2 tasks into c3 task

dczulada:


rdingwell:
I thought we moved these into the c1 and c2 tests so we wouldn't have to
upload the same documents again for validation.  Why move theme back into
an isolated task?

ssayer:
@rdingwell Because if you fail the C3 validation, but not the C1/C2, it marks you as failing both C1/C2 and C3 instead of just C3.

lib/validators/qrda_cat1_validator.rb
```diff
@@ -7,11 +7,14 @@ class QrdaCat1Validator < QrdaFileValidator
 
     self.validator_type = :result_validation
 
-    def initialize(bundle, measures = [])
+    def initialize(bundle, c3_validation, measures = [])
       @measures = measures
-      @validators = [CDA.instance,
-                     Cat1R2.instance,
```
ssayer: We should be using Cat1 not Cat1R2

lib/validators/qrda_cat1_validator.rb
```diff
@@ -7,11 +7,14 @@ class QrdaCat1Validator < QrdaFileValidator
...
-                     HealthDataStandards::Validate::DataValidator.new(bundle, measures.collect(&:hqmf_id))]
+      if c3_validation
+        @validators = [HealthDataStandards::Validate::DataValidator.new(bundle, measures.collect(&:hqmf_id))]
+      else
+        @validators = [CDA.instance,
+                       Cat1R2.instance]
```
ssayer: We should be using Cat1 not Cat1R2

ssayer:
Can someone clean up the commits in this PR so i can pull it in? @jaebird123 @dczulada

PR #284, opened 2015-12-17
Force Ruby 2.2.3 on CircleCI + remove bundler-audit.

oniofchaos:
`bundler-audit` is complaining about an insecure gem version that is due to a hardcoded dependency in `health-data-standards`. I tried upgrading to use the Q-Centrix fork but that requires `rails 4.2+` which requires more substantial updates.


oniofchaos:
Whoops, meant to make this on my fork. Please ignore :smile:

PR #293, opened 2016-01-05
View Uploaded XML

jaebird123:
added uploaded xml document viewer with styling. execution errors are now displayed in a table with the ability to jump to an error in the xml if an xml location is specified. navigation bar on xml viewer allows the user to move between errors from anywhere on the page. execution errors in the xml can be displayed with a button to the right of each error


app/views/test_executions/results/_subnav.html.erb
```diff
@@ -0,0 +1,15 @@
+<div>
+  <div class = "subnav navbar">
+    <div class = "navbar-inner">
```
thecristen: Let's name this partial along with the classname with something more descriptive. XMLNav or something like that? And does it need the extra wrapper div?

app/views/test_executions/_execution_results.html.erb
```diff
@@ -44,24 +48,33 @@
...
+
+  <% # XML Viewer %>
+  <div>
+    <% error_maps = get_error_mapping(execution) %>
+    <div id = 'xml_frame'>
+      <%= render partial: "test_executions/results/subnav"%>
+      <h3>Uploaded XML</h3>
+      <% error_maps.each do |error_map| %>
+        <div>
+          <div class = 'expando file_name'>
```
thecristen: Is `expando` a typo?

app/assets/stylesheets/cypress/_test_executions.scss
```diff
@@ -31,3 +33,94 @@
...
+.navbar-fixed-top {
+  height: 3em;
+  background: $body-bg;
+  border-bottom: 1px solid $panel-default-border;
+
+  .container {
+    padding-left: 3em;
+  }
+}
+
```
thecristen: Most of these selectors need to be more specific or else they'll effect the rest of the site in ways you probably don't intend.

app/assets/javascripts/test_executions.js
```diff
@@ -5,7 +5,46 @@ ready = function() {
...
+    return false;
+  });
+
+  function show_error_popup_and_jump(href) {
+    if ($(href)) {
+      $('button.error-popup-btn').popover('hide'); /* hide all error popovers before showing the new one */
+
+      var scroll_time = 300;      /* (milisec) time it takes to scroll from one error popover to another */
+      var highlight_time = 2000;  /* (milisec) time highlight lasts for individual error in error popover */
+      var pixels_down_page = 500; /* number of pixels down the page the error will apear after scrolling */
```
thecristen: where does this number come from? Will this work ok on every browser/resolution or would it be better to calculate this height based on the heights of certain elements or somesuch?

app/assets/stylesheets/cypress/_test_executions.scss
```diff
@@ -31,3 +33,94 @@
   color: $btn-success-color;
   border-color: $btn-success-border;
 }
+
+.tab-content {
+  margin-top: 1em;
+}
+
+/* * * * * * * * * * * * * * * * * * * * */
```
thecristen: I think this chunk of CSS should be in `_xml_view.scss`

PR #294, opened 2016-01-06
only show product tests for selected certification types

dehall:


app/views/products/show.html.erb
```diff
@@ -40,14 +40,14 @@
   <div class="product-test-tabs">
     <ul>
       <% APP_CONFIG.tests.each do |c, certification_test| %>
-        <% if certification_test['certifications'] & certifications(@product).keys() %>
```
dehall: the bug was here - the `&` means set intersection, and the goal was that if there is no intersection then the product is not certifying to the given certification criteria. the problem is that no intersection returns an empty array `[]` which is a truthy value, so it entered the body of the `if` no matter what.  I extracted the fixed logic to a helper method for a little better readability

PR #298, opened 2016-01-12
Smoking bun

rdingwell:


lib/validators/calculating_smoking_gun_validator.rb
```diff
@@ -0,0 +1,82 @@
+module Validators
+  class CalculatingSmokingGunValidator < SmokingGunValidator
+    include Validators::Validator
+    def initialize(measures, records, test_id)
+      if measures.length > 1
+        # throw some kind of an error here because we only want to do this for
```
okeefm: Is there supposed to be an actual error thrown here, or is this just left as a hook for the UI people?

rdingwell: not sure what we really want to do there.  I reality it should make sure
there is only 1 measure being tested but we would need to make sure the ui
and other code could handle that

okeefm: I think `add_error` should work in this context, so could we throw an error like "Too many measures provided for a single-measure test" in the initializer? Not sure if we would then want to not validate the first measure, or continue to evaluate it anyways.

rdingwell: but the error would be because the software did something it wasn't
supposed to do not the user , so I dont think add_error is appropriate in
this case.

okeefm:
Rebased and merged

PR #300, opened 2016-01-13
changed product status display

jaebird123:
. now split into 'C1 Checklist', 'C1/C3 Measure Tests', 'C2/C3 Measure Tests', and 'C4 Filtering Tests'. each show the number of passing, failing, not started, and total tests. disabled Rails/FindEach for rubocop

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -1,4 +1,35 @@
 module ProductsHelper
+  # returns zero for all values if test is false
+  def checklist_status_values(test)
+    passing = failing = total = 0
+    return [passing, failing, total] unless test
+    passing = test.num_measures_complete
+    total = test.num_measures
+    failing = total - passing
+    [passing, failing, total]
+  end
+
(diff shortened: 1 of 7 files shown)


app/helpers/products_helper.rb
```diff
@@ -1,4 +1,35 @@
...
+    end
+    [passing, failing, not_started, total]
+  end
+
+  def filtering_test_status_values(tests)
+    passing = failing = not_started = total = 0
+
+    # add content here when C4 tasks are finalized
+
+    [passing, failing, not_started, total]
```
okeefm: There are a couple of places where we use this construct. Might it be worth making this a small class, rather than an array? @ssayer/ @rdingwell / others; opinions?

jaebird123:
screen shot 2016-01-19 at 2 21 02 pm
screen shot 2016-01-19 at 2 21 38 pm

okeefm:
Rebased and merged.

PR #303, opened 2016-01-14
Delete the C4 subclasses

schreiaj:
As well as supporting views and UI elements referencing creating them manually.


okeefm:
What roles does `FilteringTestsController` play? Should that also have been deleted, or is that staying?

okeefm:
Rebased and merged.

PR #307, opened 2016-01-20
measure list display one table

jaebird123:
changed measure list to one table. table displays columns of CMS ID, measure name, link to c1 task, and link to c2 task. each column is sortable


jaebird123:
screen shot 2016-01-14 at 11 31 37 am
screen shot 2016-01-14 at 11 31 51 am
(sorry)

PR #308, opened 2016-01-20
C1checklist

dczulada:
Checklist Tests are create for 4 measures

4 measures are selected based upon the number of interesting "Data Criteria."  Interesting data criteria are criteria with attributes or negation.  Interesting data criteria are combined with random other criteria of other types.  At most, there is 1 entry per data criteria type.

Data Criteria are searched in order set in the cat1checklist.yml.  This file is generated by a script that uses a csv file that is based on information in the PDF provided by onc.


okeefm:
Can you squash these commits into 1-2 commits, so that we don't have quite so many 1-2 line commits in the pull request?

okeefm:
Not sure quite where the Travis status went on this one.

okeefm:
Yay travis is back

Are there UI changes in this commit? If so, screenshots would be helpful

dczulada:
Checklist UI updated to remove Valueset Ids and column added for the required attributes
screen shot 2016-01-28 at 1 22 42 pm

PR #309, opened 2016-01-20
making QRDA schematron errors warnings

rdingwell:


app/helpers/products_helper.rb
```diff
@@ -22,7 +22,7 @@ def measure_test_status_values(tests, task_type)
     [passing, failing, not_started, total]
   end
 
-  def filtering_test_status_values(tests)
+  def filtering_test_status_values(_tests)
```
okeefm: Why'd this get changed to `_tests`?

rdingwell: Rubocop did that.

okeefm: Ok, that works.

PR #313, opened 2016-01-21
Add browser tests

thecristen:
Surprise!

With this change you can run your Cucumber tests in a browser. This can be helpful in trying to understand what on earth your test is actually doing.

For example,
```
IN_BROWSER=true PAUSE=1 bundle exec cucumber -n "Successful Create Vendor"
```
The `PAUSE` adds time between the steps.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ group :development, :test do
   gem 'scss_lint', require: false
   gem 'capybara'
   gem 'capybara-accessible'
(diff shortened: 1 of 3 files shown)


thecristen:
Travis is failing because overcommit updated... is there a way to get travis to `overcommit --sign`? Not sure how else to approach that.

okeefm:
I'm testing to see if I make a PR with another commit on the branch (after running `overcommit --sign`) if that will take care of it. See #315

PR #316, opened 2016-01-26
Record views modified to display patient lists for tasks and product …

dczulada:
…tests

Button added to test executions to view patients

### DIFF
--- a/app/controllers/records_controller.rb
+++ b/app/controllers/records_controller.rb
@@ -1,6 +1,5 @@
 class RecordsController < ApplicationController
   before_action :set_record_source, only: [:index, :show, :by_measure]
-  add_breadcrumb 'Master Patient List', :records_path
 
   def download_full_test_deck
     product = Product.find(params[:id])
@@ -35,15 +34,20 @@ class RecordsController < ApplicationController
     if params[:bundle_id]
       @bundle = Bundle.find(params[:bundle_id])
       @source = @bundle
+      add_breadcrumb 'Master Patient List', :records_path
     elsif params[:product_test_id]
       @product_test = ProductTest.find(params[:product_test_id])
       @source = @product_test
+      add_breadcrumb 'Test: ' + @product_test.name, "/products/#{@product_test.product.id}/product_tests/#{@product_test.id}"
+      add_breadcrumb 'Patient List', records_path(product_test_id: @product_test.id)
     elsif params[:task_id]
       @task = Task.find(params[:task_id])
       @source = @task
+      add_breadcrumb 'Patient List', records_path(task_id: @task.id)
     else
       @bundle = Bundle.first
       @source = @bundle
+      add_breadcrumb 'Master Patient List', :records_path
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/views/records/_records_list.html.erb
```diff
@@ -31,7 +31,13 @@
...
       <% @records.each do |r| %>
         <tr>
-          <td><%= link_to full_name(r), record_path(:id => r.id), class: 'navbar-item' %></td>
+          <% if @task %>
+            <td><%= link_to full_name(r), record_path(:id => r.id, :task_id => @task.id), class: 'navbar-item' %></td>
+          <% elsif @product_test %>
+            <td><%= link_to full_name(r), record_path(:id => r.id, :product_test_id => @product_test.id), class: 'navbar-item' %></td>
+          <% else %>
+            <td><%= link_to full_name(r), record_path(:id => r.id), class: 'navbar-item' %></td>
+          <% end %>
```
thecristen: it looks like i'd left an erroneous `.navbar-item` class there... would you mind removing it

app/controllers/records_controller.rb
```diff
@@ -35,15 +34,20 @@ def set_record_source
...
+      add_breadcrumb 'Master Patient List', :records_path
     elsif params[:product_test_id]
       @product_test = ProductTest.find(params[:product_test_id])
       @source = @product_test
+      add_breadcrumb 'Test: ' + @product_test.name, "/products/#{@product_test.product.id}/product_tests/#{@product_test.id}"
+      add_breadcrumb 'Patient List', records_path(product_test_id: @product_test.id)
     elsif params[:task_id]
       @task = Task.find(params[:task_id])
       @source = @task
+      add_breadcrumb 'Patient List', records_path(task_id: @task.id)
```
thecristen: it seems like it might be appropriate to provide some information to the user about the chosen task here?

okeefm:
Similar to the other one, if you could provide screenshots of what it should look like, that would be awesome

app/controllers/records_controller.rb
dczulada: I don't think there are any show pages for a task and all tasks share the same patient list as the product test.  Might make sense to remove the ability to filter patients on task.

dczulada:
"View Patients" is a button under the Download Patients button 
screen shot 2016-01-28 at 1 24 59 pm

dczulada:
The Patient List has a breadcrumbs back to the test
screen shot 2016-01-28 at 1 26 56 pm

app/controllers/records_controller.rb
```diff
@@ -35,15 +34,16 @@ def set_record_source
...
       @bundle = Bundle.find(params[:bundle_id])
       @source = @bundle
+      add_breadcrumb 'Master Patient List', :records_path
     elsif params[:product_test_id]
       @product_test = ProductTest.find(params[:product_test_id])
       @source = @product_test
-    elsif params[:task_id]
-      @task = Task.find(params[:task_id])
-      @source = @task
+      add_breadcrumb 'Test: ' + @product_test.name, "/products/#{@product_test.product.id}/product_tests/#{@product_test.id}"
```
okeefm: Is there a reason this line kind of mixes string interpolation and string concatenation?

ssayer: Also we shouldn't be constructing URLs by hand, Use the URL/Path helpers generated by Rails

test/controllers/records_controller_test.rb
```diff
@@ -31,14 +31,6 @@ class RecordsControllerTest < ActionController::TestCase
     assert assigns(:product_test)
   end
 
-  test 'should get index scoped to task' do
```
okeefm: Why did this get removed?

okeefm:
One last note; otherwise looks good to me.

PR #319, opened 2016-01-28
Cypress3 measure eval

okeefm:
Adds back the Measure Evaluation code from Cypress 2.x, with modifications to fit the Cypress 3.x architecture. Run `bundle exec rake evaluator:evaluate_all` to generate C1, C2 and C3 tests for every measure, and to execute them with a QRDA Category 3 and a zip of QRDA Category 1s. Run `bundle exec rake evaluator:cleanup` to remove those tests.

~~Of note: in order for the calculator to work properly, you need to have the 2.7.0 bundle in your `bundles/` folder, labeled as `bundle-latest.zip`, so the calculator can install it.~~ Update 1 Feb 2016: This is no longer true with the rewritten `lib/cypress/cat_3_calculator` file used by the rake task. Still true with the `script/cat_3_calculator.rb` file.

The random segmentation faults I was experiencing with this code earlier this week seem to have disappeared after an uninstall and reinstall of MongoDB, so... ¯\\_(ツ)_/¯

`lib/measure_evaluator.rb` is `#:nocov:`'d out because it's a test file, so including it in code coverage makes no sense.


okeefm:
`Brakeman` is sad about the `eval` call to `cat_3_calculator.rb`, so I may refactor the calculator to work more like the `CalculatingSmokingGunValidator` does now and move it into `lib/`.

PR #320, opened 2016-01-28
C4

jaebird123:
adds c4 certification. filtering tests with child cat1 filter task and cat3 filter task are generated on product creation. filter test status is shown on product show page in main product status display and on CQM Filtering tab. added selection of measure for c4 filtering tests. added associated model, helper, and integration tests


okeefm:
Can you provide screenshots for the visual changes from this pull request?

jaebird123:
screen shot 2016-01-28 at 1 30 58 pm
screen shot 2016-01-28 at 1 32 17 pm

tstrass:
Don't merge yet. I think we still need a change from @jaebird123

thecristen:
Just needs a rebase from someone. Might be easier to do if you squash this first.

okeefm:
Also has merge conflicts which will have to get resolved in the rebase.

app/models/filtering_test.rb
```diff
@@ -1,3 +1,118 @@
...
+    provider = Provider.find(record.provider_performances.first['provider_id'])
+    addresses = []
+    provider.addresses.each do |address|
+      addresses << { 'street' => address.street, 'city' => address.city, 'state' => address.state, 'zip' => address.zip,
+                     'country' => address.country }
+    end
+    { 'npis' => [provider.npi], 'tins' => [provider.tin], 'addresses' => addresses }
+  end
+
+  def lookup_problem
```
dehall: One concern here is that a measure may have multiple diagnosis criteria in a union so it's possible that no records might have a given diagnosis. I think we want to ensure that we always pick a diagnosis that some record actually has. Not sure what the best way to do that is though. Loop & test each possibility?  

Also that raises the question is it possible for a set of records to not have any diagnosis on them when there is a diagnosis criteria? (for example, if the only diagnosis criteria in the measure is unioned with an encounter or something)

app/models/product.rb
```diff
@@ -71,17 +71,50 @@ def measure_ids
...
+  end
+
+  def add_filtering_tests(measure)
+    save
+    reload_relations
+    filtering_tests = product_tests.select { |product_test| product_test.is_a? FilteringTest }
+    if filtering_tests.count == 0
+      criteria = %w(races ethnicities genders payers).shuffle
+      filter_tests = []
+      filter_tests << build_filtering_test(measure, criteria.shift(2))
```
dczulada: shift removes criteria from the array.  This will throw a null pointer if you get a measure that doesn't have problem and you need to use the criteria for a third demographic filter consider using the following 

criteria[0,2] which will select 2 values starting at the 0 index
criteria[2,2] which will select 2 values starting at the 2 index

app/views/products/_filter_test_link.html.erb
```diff
@@ -0,0 +1,16 @@
...
+<% if task.passing? %>
+  <i class = 'fa fa-fw fa-check text-success' aria-hidden="true"></i>
+  <%= link_to 'view', "/tasks/#{task.id}/test_executions/new" %>
+<% elsif task.failing? %>
+  <i class = 'fa fa-fw fa-times text-danger' aria-hidden="true"></i>
+  <%= link_to 'retry', "/tasks/#{task.id}/test_executions/new" %>
+<% else %>
+  <i class = 'fa fa-fw fa-play-circle text-info invisible' aria-hidden="true"></i>
+  <%= link_to 'start', "/tasks/#{task.id}/test_executions/new" %>
+<% end %>
```
jaebird123: should use rails routes

app/views/test_executions/_execution_status_table.html.erb
```diff
@@ -11,19 +21,21 @@
       <th scope='col' class = 'col-sm-1'>Status</th>
       <th scope='col' class = 'col-sm-2'>QRDA Errors</th>
       <th scope='col' class = 'col-sm-2'>Reporting Errors</th>
-      <% if task.product_test.product.c3_test %>
-        <th scope='col' class = 'col-sm-2'>Submission Errors</th>
+      <% if task.product_test.product.c3_test && viewing_measure_test %>
+        <td scope='col' class = 'col-sm-2'>Submission Errors</td>
```
jaebird123: supposed to be th?

lib/ext/measure.rb
```diff
@@ -14,7 +14,11 @@ def display_name
   end
 
   def cms_int
-    cms_id[3, cms_id.index('v') - 3].to_i
+    return 0 unless cms_id
+    start_marker = 'CMS'
+    end_marker = 'v'
+    cms_id[/#{start_marker}(.*?)#{end_marker}/m, 1].to_i
+    # cms_id[3, cms_id.index('v') - 3].to_i
```
jaebird123: remove commented out code

app/models/product.rb
tstrass: That was changed a while ago. Not sure how it got back to a shift. Probably in one of the rebases. I'll change it back.

app/models/filtering_test.rb
rdingwell: I have the feeling that there is a good possibility of this causing nil pointer errors at some point in time.  If the measure has no diagnosis or there are no records with a diagnosis from the measure the last line in the method will cause a nil pointer.

tstrass: How should we handle those cases?

app/models/filtering_test.rb
```diff
@@ -1,3 +1,123 @@
...
+    measure = Measure.find_by(hqmf_id: measure_ids.first)
+    criteria_keys = measure.hqmf_document.source_data_criteria.keys
+    # determine which data criteira are diagnoses, and make sure we choose one that one of our records has
+    diagnosis_criteria_key = criteria_keys.find do |key|
+      criteria_def = measure.hqmf_document.source_data_criteria.send(key).definition
```
rdingwell: any particular reason send is being used here instead of just accessing the hash [] ?

app/models/product.rb
```diff
@@ -71,17 +71,50 @@ def measure_ids
...
-      product_tests.in(measure_ids: old_measure_id).destroy
+    to_remove_ids.each { |old_measure_id| product_tests.in(measure_ids: old_measure_id).destroy }
+
+    add_filtering_tests(ApplicationController.helpers.pick_measure_for_filtering_test(untouched_ids + new_ids)) if c4_test
+  end
+
+  def add_filtering_tests(measure)
+    save
+    reload_relations
+    filtering_tests = product_tests.select { |product_test| product_test.is_a? FilteringTest }
```
rdingwell: We can probably trim this down a bit by rewriting this to just count the number of filter tests.  product_tests.count(_type: "FilterTest") .  Could/should create scoped query on product to avoid selecting and filtering test types .

app/models/product_test.rb
```diff
@@ -30,7 +32,7 @@ class ProductTest
   mount_uploader :patient_archive, PatientArchiveUploader
 
   after_create do |product_test|
-    ProductTestSetupJob.perform_later(product_test)
+    ProductTestSetupJob.perform_later(product_test) unless product_test.is_a? FilteringTest
```
rdingwell: I know why this is being done.  The filtering test used to build the tests, generate the  records and then save the tests.  This allowed the records to be created before the test was actually created and saved and allowed this callback to do its job.  Now however that is not the case and having this callback seems only meaningful for MeasureTest so maybe it should be moved there.

app/views/products/_filter_test_link.html.erb
```diff
@@ -0,0 +1,16 @@
+<%
+
+# local variable 'task'
+
+%>
+
+<% if task.passing? %>
+  <i class = 'fa fa-fw fa-check text-success' aria-hidden="true"></i>
+  <%= link_to 'view', "/tasks/#{task.id}/test_executions/new" %>
```
rdingwell: should probably use rout helpers for the urls here.

app/views/products/_filter_test_link.html.erb
```diff
@@ -0,0 +1,16 @@
+<%
+
+# local variable 'task'
+
+%>
+
+<% if task.passing? %>
```
rdingwell: Not a huge deal seeing how its a small amount of code but might consider revising to use a helper of some sort for this.  All that seems to be changing are a couple of styles and the link name

app/views/products/_filter_test_link.html.erb
rdingwell: yes you should use the rout helpers.   The entire code base should be gone through at some point to refactor any hard coded links like this.  Not going to mention any of the others I see here.

app/views/products/show.html.erb
```diff
@@ -74,8 +74,9 @@
         <% elsif c == 'FilteringTest' %>
           <% # This test should be a no-op but leaving it there until we know the tests are generated %>
           <% if @product.product_tests.filtering_tests.exists? %>
-
-            The <%= certification_test['title'] %> is <%= @product.product_tests.filtering_tests.first.status %>.
+            <%= render partial: 'filtering_test_status_display', locals: { tests: @product.product_tests.filtering_tests } %>
```
rdingwell: Remove the if else block and only render the if portion

lib/ext/measure.rb
```diff
@@ -14,7 +14,10 @@ def display_name
   end
 
   def cms_int
```
rdingwell: What is this used for?  Also, if there is more than 1 measure without a cms id then they all will be the same cms_int "0"

thecristen: This integer's mostly used to sort lists of measures. We shouldn't have measures without CMS IDs, but if we do, they'll just all get grouped together in sorting...

PR #321, opened 2016-01-28
Navbar stays on Dashboard for Test Patient List

dczulada:
Previously the navbar highlighted the Master Patient List
screen shot 2016-01-28 at 4 07 15 pm

### DIFF
--- a/app/views/application/_navbar.html.erb
+++ b/app/views/application/_navbar.html.erb
@@ -9,13 +9,13 @@
     <div class="collapse navbar-collapse" id="cypressNavbar">
       <ul class="nav navbar-nav navbar-right">
         <% if user_signed_in? %>
-          <li <% if params[:controller] != 'records' && params[:controller] != 'registrations' %>class="active"<% end %>>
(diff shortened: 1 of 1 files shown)


thecristen:
Good catch!

app/views/application/_navbar.html.erb
```diff
@@ -9,13 +9,13 @@
     <div class="collapse navbar-collapse" id="cypressNavbar">
       <ul class="nav navbar-nav navbar-right">
         <% if user_signed_in? %>
-          <li <% if params[:controller] != 'records' && params[:controller] != 'registrations' %>class="active"<% end %>>
+          <li <% if (params[:controller] == 'records' && params[:product_test_id]) || (params[:controller] != 'records' && params[:controller] != 'registrations') %>class="active"<% end %>>
```
jaebird123: maybe the long conditional could go in a helper function

ssayer: you should be using the rails routes, not looking at params to figure out username controller

okeefm: http://stackoverflow.com/a/4354547 is a decent way to use the Rails routes to do just this, e.g. `current_page?(:some_route)`

PR #324, opened 2016-01-29
New color scheme + more

thecristen:
I changed the color scheme to something that I think looks more elegant and less bright. Feel free to comment.

Also checked this in all the browsers I know of. Looks good back to IE10 on Win 7. Is this sufficient?

Also added even more accessibility checks. Supposedly...
- This PR passes 508
- This PR passes WCAG 2.0 AA. Going above and beyond! 10X!

Also did two more chores:
- Changed `.erb` files to `.html.erb`
- Turned off autocomplete for product name and vendor name

## Screenshot parade

screen shot 2016-01-29 at 4 15 08 pm

screen shot 2016-01-29 at 4 13 20 pm
screen shot 2016-01-29 at 4 13 04 pm
The next page still needs work, for another PR...

screen shot 2016-01-29 at 4 12 48 pm


app/assets/javascripts/records.js
```diff
@@ -10,7 +10,14 @@ ready = function() {
     } else {
       $.get('records/by_measure');
     }
+  }).on('open opening closing close select selecting unselect unselecting', function(e){
+    console.log(e);
```
jaebird123: maybe leave console.log out

PR #326, opened 2016-02-01
Added C3 Cat1 and C3 Cat3 to Product Status Display

jaebird123:
separated (c1,c3) and (c2,c3) measure tests into c1, c3 cat1, c2, c3 cat3 for product status display.
fixed issue with measure_tests table not reporting the status of c3 test executions (used to just report status of c1 or c2).

fixed issue on vendor show page. spacer " | " was showing up in inappropriate situations

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -9,17 +9,16 @@ module ProductsHelper
     [passing, failing, total]
   end
 
-  def measure_test_status_values(tests, task_type)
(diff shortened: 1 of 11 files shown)


jaebird123:
screen shot 2016-02-03 at 12 09 47 pm
screen shot 2016-02-03 at 12 10 06 pm
screen shot 2016-02-03 at 12 09 06 pm

PR #329, opened 2016-02-02
require that product has c1 or c2 selected

dehall:
c1_or_c2


ssayer:
Can you change the error message to "Product must certify to at least C1 or C2"?

dehall:
new_msg

PR #330, opened 2016-02-05
fix the model used for provider filtering

dehall:
NPI and TIN should be in cda_identifiers, not their own individual fields


rdingwell:
Is this pulled into the the C4 pr ?

dehall:
Not yet. This issue existed before the c4 branch plus I thought it would be quicker to pull this separately and have the C4 branch rebase.

PR #331, opened 2016-02-09
Bundle upload in UI

dehall:
Ability to upload bundle in UI. Note there is no link to the page yet (because we don't have an admin UI yet). Until then manually go to /bundles

Screenshots:

Add bundle:
add_bundle

When there is a single bundle you can delete it:
single_bundle

When there is more than one bundle you can change the default or delete one of the non-default bundles:
multiple_bundles


ssayer:
We'll need a setting in config/cypress.yml to allow us to turn this page off until the admin system is ready.

dehall:
bundles page is readonly when disabled:
disabled_page

attempting to call any actions directly (ex. POSTing directly to the endpoint) will just redirect back to the main bundles list

PR #337, opened 2016-02-11
Added :timeoutable to user, and set it to time out after 2 hours

okeefm:


okeefm:
I don't know why Travis didn't run on this...

PR #339, opened 2016-02-11
Bumped the default bundle version to 2.9.9, and changed around the ef…

okeefm:
Bump the default version to `2.9.9`, and change the measure evaluator effective date stuff to be more right (effective date wasn't ending within the correct year).


okeefm:
Stand by on this PR; one more fix coming to support the new record randomization flag.

PR #340, opened 2016-02-12
ensure that deleting a bundle deletes all the associated pieces

dehall:
associated pieces being measures, patient cache, records.

note: initializers/mongo.rb came from cypress2 and needed some cleanup but there is probably more that could be done - open to suggestions

### DIFF
--- a/config/initializers/mongo.rb
+++ b/config/initializers/mongo.rb
@@ -1,0 +1,17 @@
+module QME
+  patient_cache = Mongoid.default_client['patient_cache']
+  # create a unique index for patient cache, this prevents a race condition where the same patient can be entered multiple times for a patient
+  patient_cache.indexes.create_one({ 'value.measure_id' => 1, 'value.sub_id' => 1, 'value.effective_date' => 1, 'value.patient_id' => 1 },
+                                   'unique' => true)
+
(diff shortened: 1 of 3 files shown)


config/initializers/mongo.rb
```diff
@@ -0,0 +1,17 @@
+module QME
+  patient_cache = Mongoid.default_client['patient_cache']
+  # create a unique index for patient cache, this prevents a race condition where the same patient can be entered multiple times for a patient
+  patient_cache.indexes.create_one({ 'value.measure_id' => 1, 'value.sub_id' => 1, 'value.effective_date' => 1, 'value.patient_id' => 1 },
```
okeefm: I can't find any information on `create_one` in Mongoid's documentation or on rubydoc.info; what's the difference between it and `create`? @rdingwell may be a better person to ask

dehall: it was `create` in the original from cypress v2, but it threw an error `undefined method 'create' for #<Mongo::Index ...`
I changed it to `create_one` after seeing it here: https://docs.mongodb.org/ecosystem/tutorial/ruby-driver-tutorial/#indexing

okeefm: Ok, that works.

PR #358, opened 2016-02-19
adding a single duplicate record to measure test archive

rdingwell:
screen shot 2016-02-22 at 5 20 07 pm

screen shot 2016-02-22 at 5 22 04 pm

### DIFF
--- a/app/views/products/_product_form.html.erb
+++ b/app/views/products/_product_form.html.erb
@@ -24,7 +24,11 @@
           <% end %>
 
           <%= f.form_group :randomize_records do %>
-            <%= f.check_box :randomize_records, label: 'Randomize Records', checked: true, disabled: params[:action] == "edit" %>
+            <%= f.check_box :randomize_records, label: 'Randomize Records', checked: @product.new_record? ? true : @product.randomize_records, disabled: !@product.new_record? %>
+          <% end %>
+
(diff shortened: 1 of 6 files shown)


app/views/products/_product_form.html.erb
```diff
@@ -24,7 +24,11 @@
           <% end %>
 
           <%= f.form_group :randomize_records do %>
-            <%= f.check_box :randomize_records, label: 'Randomize Records', checked: true, disabled: params[:action] == "edit" %>
+            <%= f.check_box :randomize_records, label: 'Randomize Records', checked: params[:action] == 'edit' ? @product.randomize_records : true, disabled: params[:action] == 'edit' %>
```
ssayer: You should use @product.new_record? instead of params[:action] == 'edit' to detect if we should disable

app/views/products/_product_form.html.erb
```diff
@@ -24,7 +24,11 @@
...
+          <% end %>
+
+          <%= f.form_group :duplicate_records do %>
+            <%= f.check_box :duplicate_records, label: 'Duplicate Records', checked: params[:action] == 'edit' ? @product.duplicate_records : true, disabled: params[:action] == 'edit' %>
```
ssayer: Ditto previous comment

PR #364, opened 2016-02-23
Cache Status

jaebird123:
added rails caching to vendor helper function get_product_status_values(). this function reports the number of passing, failing, not started, and total tests for each certification type and is used on the vendor show page for each product

### DIFF
--- a/test/helpers/vendors_helper_test.rb
+++ b/test/helpers/vendors_helper_test.rb
@@ -175,4 +175,30 @@ class VendorsHelperTest < ActiveJob::TestCase
     assert_equal '2 tests failing', status_to_display_text('Failing', 'C1', certs.C1)
     assert_equal '2 tests to go', status_to_display_text('Not Complete', 'C3', certs.C3)
   end
+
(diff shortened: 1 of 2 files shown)


test/helpers/vendors_helper_test.rb
```diff
@@ -175,4 +175,31 @@ def test_status_to_display_text
     assert_equal '2 tests failing', status_to_display_text('Failing', 'C1', certs.C1)
     assert_equal '2 tests to go', status_to_display_text('Not Complete', 'C3', certs.C3)
   end
+
+  def test_get_product_status_values_performs_caching
+    ActionController::Base.perform_caching = true
+    assert_equal false, Rails.cache.exist?("#{@product.cache_key}/status_values"), 'cache key for product should not exist before function call'
```
okeefm: Is caching enabled in the `test` environment?

jaebird123: yah good point, caching is enabled in test env. I'll take out that line

PR #366, opened 2016-02-24
Don't validate unexpected files

dehall:
Validators now have a "can_continue" function/attribute which can be used to skip the rest of the validators for the given file. This means that order of validators defined in the tasks is now relevant.

The way this is used now is to use the smoking gun validator's expected records check to bypass the rest of the validators if we get a file we don't expect. This will happen for c1, c3 cat 1, and c4 cat 1. (C1 and C4 cat 1 already had smoking gun validator so just the order changed. C3 cat 1 now has the smoking gun validator but it only performs the expected files check, not the data validation)


ssayer:
@rdingwell can you review this?

PR #368, opened 2016-02-24
C1 Manual Value Set oid

jaebird123:
added value set oid to manual entry checklist. added field value sets for types 'CD' and length_of_stay

screen shot 2016-02-24 at 1 12 39 pm

### DIFF
--- a/app/helpers/checklist_tests_helper.rb
+++ b/app/helpers/checklist_tests_helper.rb
@@ -19,4 +19,23 @@ module ChecklistTestsHelper
       return 'Fulfills - ' + data_criteria[referenced_criteria].description if data_criteria[referenced_criteria]
     end
   end
+
+  # argument fv stands for field_value (Hash)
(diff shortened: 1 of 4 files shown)


app/helpers/checklist_tests_helper.rb
```diff
@@ -19,4 +19,23 @@ def fulfills_reference(measure, referenced_criteria)
...
+    msg += 'stay'
+    if fv['high']
+      msg += less_than_symbol(fv['high']['inclusive?'])
+      msg += "#{fv['high']['value']} #{fv['high']['unit']}"
+    end
+    msg
+  end
+
+  def less_than_symbol(inclusive)
+    inclusive ? ' &#8804; ' : ' < '
```
jaebird123: right now, I used the unicode character for less than or equal to. I'm not sure if we want this or if we just want "<="

PR #376, opened 2016-02-25
Supplemental Data Error Reporting

jaebird123:
added table for supplemental data reporting errors. errors are now reported for each supplemental data (Race, Sex, Ethnicity, Payer) with mismatched expected and reported values. this adds to the total possible number of execution errors

screen shot 2016-02-25 at 4 22 40 pm

### DIFF
--- a/app/helpers/test_executions_results_helper.rb
+++ b/app/helpers/test_executions_results_helper.rb
@@ -1,0 +1,69 @@
+module TestExecutionsResultsHelper
+  def get_select_history_message(execution, is_most_recent)
+    msg = ''
+    msg << 'Most Recent - ' if is_most_recent
+    msg << execution.created_at.in_time_zone('Eastern Time (US & Canada)').strftime('%b %d, %Y at %I:%M %p (%A)')
+    msg << ' (passing)' if execution.passing?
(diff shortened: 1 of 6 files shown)


ssayer:
Screenshot plz

app/helpers/test_executions_results_helper.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def supplemental_data_errors(errors)
+    return_errors = errors.select do |err|
+      err.has_attribute?('error_details') && err['error_details'].key?('type') && err['error_details']['type'] == 'supplemental_data'
+    end
+    # sort by population key (IPP, DENOM, ...)
+    return_errors.sort do |a, b|
+      a.error_details.population_key <=> a.error_details.population_key
+    end
+  end
```
jaebird123: this is the only function I added. I spit the test_executions helper class because the file was over 100 lines

okeefm:
When I run C1 and C3, my reporting errors look like this: 
screen shot 2016-03-02 at 8 29 37 am

Presumably this is more meant for C2 and C3 (cat 3) error reporting, but do we want to implement it for C1 and C3 as well?

PR #377, opened 2016-02-25
Fixed all_records_for_product so it returns something

okeefm:
`ProductsHelper::all_records_for_product` now it uses the non-destructive versions of `sort_by` and `uniq`, so it actually returns something, rather than just modifying the `records` object in place


okeefm:
Don't merge this just yet; I'm writing a test for this method.

PR #379, opened 2016-02-29
don't re-create the patient qrda zip every download

dehall:


okeefm:
Would it make any sense to cache the archive on the product test if it isn't already built, or are we anticipating that state being rare enough to not worry about it?

dehall:
the qrda archive should theoretically always be there when the user clicks download so even the new null check might not be necessary.

 the existing code does also allow for downloading other formats but there is no way to download them as of now, we could theoretically cache those

okeefm:
It's probably a good idea to have the check for the moment. When we implement further format downloads, we can work on a better caching scheme.

PR #381, opened 2016-02-29
implementing checking of sub measures against uploaded qrda cat I doc…

rdingwell:
…uments for smoking gun testing.  Removed some fixtures that were not being used in the process to facilitate testing of the changes


test/models/record_test.rb
```diff
@@ -25,6 +25,6 @@ def test_record_should_be_able_to_find_original
 
   def test_record_should_be_able_to_find_calculation_results
     r = Record.find('51703a883054cf84390000d4')
-    assert_equal 1, r.calculation_results.count, 'record should have 1 calculated result'
+    assert_equal 2, r.calculation_results.count, 'record should have 1 calculated result'
```
okeefm: Should probably change the error text in this assertion to "record should have 2 calculated results"

PR #387, opened 2016-03-07
Job status

rdingwell:
Adding ActiveJob tracker which will be required for things like bundle upload in the background

### DIFF
--- a/app/jobs/measure_evaluation_job.rb
+++ b/app/jobs/measure_evaluation_job.rb
@@ -1,6 +1,6 @@
 class MeasureEvaluationJob < ActiveJob::Base
   queue_as :default
-
+  include Job::Status
   def perform(test_or_task, options)
     if test_or_task.is_a? ProductTest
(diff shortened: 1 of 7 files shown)


app/helpers/test_executions_results_helper.rb
```diff
@@ -23,7 +23,7 @@ def supplemental_data_errors(errors)
       err.has_attribute?('error_details') && err['error_details'].key?('type') && err['error_details']['type'] == 'supplemental_data'
     end
     # sort by population key (IPP, DENOM, ...)
-    return_errors.sort do |a, b|
+    return_errors.sort do |a, _b|
```
dehall: i think the fix here should be to replace the duplicate `a.error_details.population_key` with `b.error...` instead of setting b to ignored. this is a separate issue and unrelated to the core change though

rdingwell: I didn't touch that file so it must be something that rubocop put in there

PR #394, opened 2016-03-16
turn off autocorrect for many fields 

thecristen:
turn off autocapitalize for a few fields


thecristen:
ALSO: added more front-end validations.

screen shot 2016-03-16 at 2 08 35 pm

okeefm:
Note: this passed travis, even if it's not displaying: https://travis-ci.org/projectcypress/cypress/builds/116436723

thecristen:
ALSO: Re-adding higher password complexity.

Added validation on the devise pages too

so much validation

PR #396, opened 2016-03-17
Multi bundle support

dehall:
Adds support for multiple bundles. Products are now tied to a specific bundle. (Note: in Cypress 2.x , it was the product tests that were tied to the bundle, so there was some migration of properties) 

## Main Features:
 - Adds bundle selector to Product Creation page. If there is only 1 bundle installed then its title is displayed but there is no radio button. Switching the bundle results in the measures being reloaded by ajax.
 - Adds bundle selector to Master Patient List. Again if only 1 bundle then it is just named.
 - Moved patient list to be scoped by bundle, and /records page now redirects to /bundles/{default_bundle_id}/records. 
 - All calls to records and measures should be done by bundle. Ex, instead of `Record.where`, use `bundle.records.where` or `product_test.records.where` , etc

### Product Creation Page
With Multiple Bundles Installed:
product_creation_multi_bundle

With a Single Bundle Installed:
product_creation_single_bundle

### Product Update Page
With Multiple Bundles Installed:
product_update_multi_bundle

With a Single Bundle Installed:
product_update_single_bundle

### Master Patient List
With Multiple Bundles Installed:
mpl_multi_bundle

With a Single Bundle Installed:
mpl_single_bundle


app/controllers/checklist_tests_controller.rb
```diff
@@ -6,9 +6,7 @@ class ChecklistTestsController < ProductTestsController
   before_action :set_measures, only: [:show]
 
   def create
-    # bundle ATTRIBUTE IS NOT CHOSEN CORRECTLY. MUST FIX LATER ~ JAEBIRD
```
okeefm: Did this PR fix the attribute not being chosen correctly?

dehall: Oh I forgot about this file. Actually I don't think this class is even used anymore.  But the bundle is now tied to the product not the product test, so there's no need to specify the bundle id here.

dehall: Was thinking of another class that isn't used (filtering_test_controller). But my other statement is still relevant.

PR #399, opened 2016-03-21
API

jaebird123:
- Added json and xml responses for bundles, measures, vendors, products, product_tests, tasks, and test_executions controllers
- Added "responders" and "roar" gems
- API responds to xml and json including hyperlinks
- Added RAML api documents
- Downloadable "patients" action on product_test controller uses achive (and is faster now)
- Bulk download patients is now in the products controller (as "patients" action) instead of records controller
- Deleted unused models, views, controllers, and routes.
- Refactored product_tests controller and deleted measure_tests and filtering_tests controllers


app/views/products/_measure_selection.html.erb
```diff
@@ -1,3 +1,12 @@
+<%
+
+# local variables
+#
+#   f   (product form)
```
dehall: This partial gets re-rendered by ajax if the user changes the bundle during product creation. (see assets/javascripts/measure_selection.js : UpdateMeasureSet --> measures_controller.grouped) 

Right now this is breaking because we don't have a form object when re-rendering. We could possibly work around that but I think it would be easier to just avoid using a form object in this partial. The new version is definitely cleaner though.

app/views/products/_bulk_download.html.erb
```diff
@@ -10,7 +10,7 @@
 <% num_measure_tests_ready = product.product_tests.measure_tests.where(state: :ready).count %>
 <% if num_measure_tests_ready == num_measure_tests %>
   <p>This download contains a folder for each measure selected for this product. Inside these folders are XML documents for each patient associated with that measure.</p>
-  <%= form_for product, url: { controller: 'products', action: 'download_full_test_deck' }, :html => { :method => 'GET' } do |f| %>
+  <%= form_for product, url: { controller: 'products', action: 'patients' }, :html => { :method => 'GET' } do |f| %>to report and patients respectively.
```
dehall: i think this was an accidental copy&paste or something at the end of this line

app/controllers/bundles_controller.rb
```diff
@@ -21,6 +24,11 @@ def redirect_if_disabled
 
   def index
     @bundles = Bundle.all
+    respond_with(@bundles)
```
dehall: The index and show methods for bundles are extremely slow in the API and return a ton of data. (for example, it returns all the details for all the measures) This should probably be trimmed down somehow.

okeefm: `Bundle` could also use some indexes. There are currently none, either in Cypress or in HDS. In particular, indexes on relations (ex: `has_many :product_tests, :dependent => :destroy` could get `index: true` added to it) would probably help, in addition to paring down the returned data.

dehall:
getting the root (`/`) with accept:json or xml results in getting back the html sign-in page. not sure if that should be expected to redirect to /vendors the way it does in a browser

PR #410, opened 2016-03-30
Enabled ESLint rules

okeefm:
So it turns out that ESLint doesn't actually...you know, lint...unless you give it rules in a `.eslintrc` file. So I added one. 
Also fixed a bunch of the issues that popped up in our JS files with respect to the rules. 
Also also bumped `overcommit` to `0.32.0` in our `Gemfile.lock` because it was blocking me from committing until I did so.


app/assets/javascripts/xml_navigator.js
```diff
@@ -45,21 +45,21 @@
   }
   Navigator.prototype.prev = function() {
     this.index = this.index  - 1;
-    if(this.index == NaN || this.index < 0){this.index=0;}
+    if(this.index.isNaN || this.index < 0){this.index=0;}
```
rdingwell: where is the isNaN property coming from?  there is a javascript method isNaN(value) that will return true or false but I don't think there is an isNaN property on NaN

PR #445, opened 2016-04-27
When drilling down into an individual patient from the MPL, the heade…

dczulada:
…r navigation switched from Master Patient List to Dashboard

### DIFF
--- a/app/views/application/_navbar.html.erb
+++ b/app/views/application/_navbar.html.erb
@@ -10,10 +10,10 @@
     <div class="collapse navbar-collapse" id="cypressNavbar">
       <ul class="nav navbar-nav navbar-right">
         <% if user_signed_in? %>
-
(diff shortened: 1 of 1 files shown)


app/views/application/_navbar.html.erb
```diff
@@ -10,10 +10,10 @@
     <div class="collapse navbar-collapse" id="cypressNavbar">
       <ul class="nav navbar-nav navbar-right">
         <% if user_signed_in? %>
-
           <% curr_page = if current_page?(edit_user_registration_path)
                            :account
-                         elsif (controller.controller_name == 'records' && params[:bundle_id]) || current_page?(:records)
+                         elsif (controller.controller_name == 'records' && (params[:bundle_id] || params[:task_id].nil?)) ||
```
ssayer: Can you move this massive if statement into the application helper?

PR #457, opened 2016-05-05
Fixed bug where async product test status updates stop after hitting …

tstrass:
…back button

turbolinks ❌


ssayer:
Need to fix this by hooking into turbolinks

PR #479, opened 2016-05-19
Files for building Cypress with Packer

okeefm:


ssayer:
Need to bump version in cypress.yml to alpha6 too...can you add that to this PR?

PR #483, opened 2016-05-19
valuesets are also needed for negations

dczulada:
screen shot 2016-05-19 at 2 24 35 pm


okeefm:
This was an actual Travis failure: looks like you have some hard tabs in `app/views/checklist_tests/_field_values.html.erb` on lines 13, 14 and 17.

PR #485, opened 2016-05-23
age filter

dehall:
Adds an "Age" filter as another potential filter on demographic C4 tests. Most of the logic to handle age filtering was already there.

age_filter1

new_age_filter


dczulada:
Age 78 @ timestamp

PR #491, opened 2016-06-03
Add a page:load event and remove a data-no-turbolink

tstrass:
- For admin, add a `page:load` event
- Also in admin, remove `data-no-turbolink`

These changes were made while removing `(document).ready` events throughout the app, but then we realized we actually need them, as per [this stack overflow post](http://stackoverflow.com/questions/18770517/rails-4-how-to-use-document-ready-with-turbo-links).

### DIFF
--- a/app/assets/javascripts/admin.js
+++ b/app/assets/javascripts/admin.js
@@ -24,7 +24,8 @@ var assignmentsReady = function(){
 
 }
 
-$(document).ready(assignmentsReady)
+$(document).ready(assignmentsReady);
+$(document).on('page:load', assignmentsReady);
 
 // Place all the behaviors and hooks related to the matching controller here.
 // All this logic will automatically be available in application.js.
--- a/app/views/admin/show.html.erb
+++ b/app/views/admin/show.html.erb
@@ -4,8 +4,8 @@
   #   banner_message, smtp_settings, mode, mode_settings, debug_features
 %>
 
-<div data-no-turbolink class="settings-tabs">
-  <ul data-no-turbolink>
+<div class="settings-tabs">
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/admin.js
```diff
@@ -24,7 +24,8 @@ var assignmentsReady = function(){
 
 }
 
-$(document).ready(assignmentsReady)
+$(document).ready(assignmentsReady);
+$(document).on('page:load', assignmentsReady);
```
okeefm: Is it going to cause issues if this function is run more than once? I.E. when the `$(document).ready` fires, and when the `page:load` fires?

tstrass: @okeefm sorry I missed your comment. It is impossible for those two events to both fire on the same page. This is confirmed in the comments on the top answer on the Stack Overflow post that is linked in the PR description

thecristen: I think the concern was not about them firing at the same time, but rather firing again and (re)affecting the same DOM elements. Is it the case that, on next firing, the body is thrown out and re-inserted, so any DOM changes are happening on "new" elements?

tstrass: > on next firing

I'm sure what exactly you mean by this. Which event? To be perfectly clear:
```javascript
// turbolinks disabled
$(document).ready(func)
```
is equivalent to 
```javascript
// turbolinks enabled
$(document).ready(func); // only fires if turbolinks does not have this page cached
$(document).on('page:load', func); // fires if turbolinks has the page cached
```

This is the desired behavior in this case. @thecristen

PR #492, opened 2016-06-03
Require entry of data for all attributes. Coded, Result, Date Times, …

dczulada:
…Cumulative Medication Durations and Length of Stay.

screen shot 2016-06-03 at 9 36 57 am

### DIFF
--- a/app/models/checklist_source_data_criteria.rb
+++ b/app/models/checklist_source_data_criteria.rb
@@ -9,18 +9,32 @@ class ChecklistSourceDataCriteria
   field :source_data_criteria, type: String # this is the name of the source_data_criteria
 
   field :completed, type: Boolean
+  field :recorded_result, type: String
   field :code, type: String
   field :attribute_code, type: String
 
   validate :code_matches_valueset?
   validate :attribute_code_matches_valueset?
+  validate :result_completed?
+
+  def result_completed?
+    if recorded_result && completed
+      errors.add(:recorded_result, 'Value must be entered.') if recorded_result == ''
+    end
+  end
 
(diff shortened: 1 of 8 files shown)


dczulada:
screen shot 2016-06-03 at 3 03 24 pm

app/models/checklist_source_data_criteria.rb
```diff
@@ -9,27 +9,56 @@ class ChecklistSourceDataCriteria
...
+
+  def complete?
+    self.completed = if code.blank? && attribute_code.blank? && recorded_result.blank?
+                       nil
+                     else
+                       code_complete != false && attribute_complete != false && result_complete != false
+                     end
+  end
+
+  def result_completed?
```
okeefm: Is there a reason we're putting this `Boolean` value into a field, rather than just using this function?

app/models/checklist_source_data_criteria.rb
```diff
@@ -9,27 +9,56 @@ class ChecklistSourceDataCriteria
...
+                   [criteria[:value].code_list_id]
+                 else
+                   [criteria.negation_code_list_id]
+                 end
+      self.attribute_complete = code_in_valuesets(valueset, attribute_code)
     end
   end
 
   def code_matches_valueset?
     # validate if an code is required, and the completed box is checked
```
okeefm: This comment isn't strictly accurate any more (there is no more checkbox)

thecristen:
will these changes allow people to save a Manual Entry test that isn't completely passing? or is that outside the scope of this PR?

dczulada:
Yes, all content is now saved, even if it isn't passing

Issue #493, opened 2016-06-03
Cypress getting crashed while uploading QRDA III


sudhir-nair:
It is giving following error.
 Server Error
Something went wrong with this request. This is likely a problem with Cypress rather than a mistake on your part. Please find attached xml file just rename it to .xml.

File is a QRDA III xml for CMS111

[QRDA3 (1).txt](https://github.com/projectcypress/cypress/files/297724/QRDA3.1.txt)

ssayer:
Sudhir,

Can you confirm you're on the latest release (Alpha 6)?

-Sam

sudhir-nair:
Yes am using Alpha 6

sudhir-nair:
Am unable to retry upload of QRDA III file now.Seems like i need to create a new product and reprocess complete data.

ssayer:
Sudhir,

If you’re going to lose a non-trivial amount of work, we can try to find a workaround once we’re done testing your file.

-Sam

sudhir-nair:
It's fine...I can reprocess. Just let me know what went wrong.

dczulada:
Thanks for pointing this out.  There is a bug in Cypress that is preventing displaying the error message related to the file you uploaded.

The error in your file is related to the OBSERV id.  The file you attached is still using the id "b9352e6d-c82d-424a-a1d3-05a1f3fbaed3" from the measures released in 2014.  This id needs to be updated to "C71EF921-C721-477D-AB8A-14FB9A19C50D."  I've uploaded a file making that change which should now work.

[QRDA3.1.txt](https://github.com/projectcypress/cypress/files/297843/QRDA3.1.txt)

ssayer:
Sudhir,

This should be fixed in the latest beta1 release. If it is not, please reopen

PR #494, opened 2016-06-03
population_errors should check if its an empty array

dczulada:
population_errors.first.message was giving a 500 error when the array was empty

Fix for #493

### DIFF
--- a/app/views/test_executions/results/_supplemental_data_error_table.html.erb
+++ b/app/views/test_executions/results/_supplemental_data_error_table.html.erb
@@ -15,7 +15,7 @@
       <tr>
         <th colspan = '5' scope = 'col' class = 'col-sm-2'>
           <% if pop_error_value %>
-            <% unless population_errors.empty? %>
+            <% unless population_errors.blank? %>
               <%= population_errors.first.message %>. The following errors were also identified for this population.
             <% end %>
           <% else %>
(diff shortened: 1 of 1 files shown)


app/views/test_executions/results/_supplemental_data_error_table.html.erb
```diff
@@ -15,7 +15,7 @@
       <tr>
         <th colspan = '5' scope = 'col' class = 'col-sm-2'>
           <% if pop_error_value %>
-            <% if population_errors %>
+            <% unless population_errors.empty? %>
```
okeefm: can `population_errors` ever be `nil`? If it is, the second case will blow up, because `nil` doesn't respond to `.empty?`. Maybe we need something like `if population_errors && !population_errors.empty?`

app/views/test_executions/results/_supplemental_data_error_table.html.erb
```diff
@@ -15,7 +15,7 @@
       <tr>
         <th colspan = '5' scope = 'col' class = 'col-sm-2'>
           <% if pop_error_value %>
-            <% if population_errors %>
+            <% unless population_errors.blank? %>
```
okeefm: It doesn't look like either `Array` or `nil` have a `.blank?` method

dczulada:
screen shot 2016-06-15 at 12 36 22 pm

PR #495, opened 2016-06-06
Re-added cypress rake tasks for cleaning up the database.

rbclark:
At some point the rake tasks used to cleanup the database for the demo server was removed. This pull request adds them back.


okeefm:
Looks like there were some merge conflicts that appeared in this commit.

okeefm:
Yeah, looks like the only errors we're getting now are some weird Travis-only errors, so we'll try to get those to pass then pull it in.

PR #497, opened 2016-06-09
Renovate the splash page

tstrass:
Splash (sign-in) page renovation, following @pkroft's mockup.

**Full width:**
screen shot 2016-06-09 at 10 31 59 am

**Narrow:**
screen shot 2016-06-09 at 10 34 00 am


app/views/devise/sessions/new.html.erb
```diff
@@ -1,22 +1,69 @@
-<div class="panel panel-default">
-  <div class="panel-heading">
-    <h1 class='panel-title'>Sign In</h1>
-  </div>
+<div class="splash-panel">
+  <div class="splash-title-container">
+    <h1>cypress</h1>
+    <h2>Rigorouts &amp; repeatable testing of Electronic Health Records</h2>
```
thecristen: Rigorous

thecristen:
The sign up/forgot password/instructions links really should be next to/below the sign-in form, not at the bottom of the whole page.

app/assets/javascripts/sessions.js
```diff
@@ -0,0 +1,18 @@
+var placeSignInPanel = function() {
+  if ($('.splash-title-container').length)  {
+    var titleContainerRightPos = $('.splash-title-container').offset().left + $('.splash-title-container').width()
+    var splashPanelRightPos = $('.splash-panel').offset().left + $('.splash-panel').width()
+    var roomForSignInPanel = (splashPanelRightPos - titleContainerRightPos) > $('.sign-in-panel').width();
+    if ($(".sign-in-panel").parent().is(".splash-panel") && !roomForSignInPanel) {
+      $(".sign-in-panel").prependTo(".splash-info-panels-container");
+      $(".sign-in-panel").css({"float":"none","margin":"auto","margin-bottom":"5px","display":"block","left":"0px","position":"relative"});
```
thecristen: Could just add a class here and specify these properties in the `_sessions.scss`

app/views/devise/sessions/new.html.erb
ssayer: I'll fix this when I update the text

app/views/devise/sessions/new.html.erb
```diff
@@ -1,22 +1,69 @@
...
+    <p>Powered by Super Duper Testing Labs</p>
```
thecristen: Should the `Super Duper Testing Labs` bit be supplied through an admin setting or some such instead of hard coded?

app/views/devise/sessions/new.html.erb
```diff
@@ -1,22 +1,69 @@
...
+  </div>class="
```
thecristen: extra `class="` text be lurking

app/views/devise/sessions/new.html.erb
ssayer: Yeah that shouldn't be hardcoded, I'll remove it in my commit. I'll get feedback from the ATLs and see if what they want customized. We can make that it's own feature in Pivotal as there might be other customization they want.

thecristen:
I don't think you need any of the javascript if you use Bootstrap's responsive classes...

```erb
<div class="jumbotron splash">
  <div class="container-fluid">
    <div class="col-md-8 splash-title">
      <h1>cypress</h1>
      <p>Rigorouts &amp; repeatable testing of Electronic Health Records</p>
      <small>Powered by Super Duper Testing Labs</small>
    </div>
    <div class="col-md-4 splash-sign-in">
      <div class="panel panel-default">
        <%= bootstrap_form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
          <div class="panel-body">
            <%= f.email_field :email, autocapitalize: 'off', autocorrect: 'off', autofocus: true %>
            <%= f.password_field :password, autocomplete: "off", autocapitalize: 'off', autocorrect: 'off' %>

            <% if devise_mapping.rememberable? -%>
              <%= f.check_box :remember_me %>
            <% end -%>
          </div>

          <div class="panel-footer">
            <%= f.primary "Sign in" %>
          </div>
        <% end %>
      </div>
    </div>

  </div>
</div>
```

ssayer:
@thecristen I just tried to put the devise links in the sing up box as is, but it's going to need some design TLC

tstrass:
@thecristen only issue with using those responsive classes is that we have a fixed height image as our background. It would look bad if we stretched it. I was doubting I'd be able to find bootstrap classes that did exactly what I wanted, but let me know if you think of a way to do it without javascript.

thecristen:
@tstrass The responsive classes were more for handling the sign in panel and the text, not the div with the image.

tstrass:
@thecristen I understand, but the panel goes in the div with the image. It needs to either neatly placed over the image or completely outside of it. It would look bad otherwise

PR #500, opened 2016-06-14
Sync default bundle between app and cypress.yml

tstrass:


app/controllers/admin_controller.rb
```diff
@@ -20,4 +20,16 @@ def show
   def require_admin
     raise CanCan::AccessDenied.new, 'Forbidden' unless current_user.has_role? :admin
   end
+
+  def sub_yml_setting(key, val)
```
okeefm: Not saying it's wrong, but how come `sub_yml_setting` got moved here from `SettingsController`?

tstrass: @okeefm I needed to call that function [here](https://github.com/projectcypress/cypress/pull/500/files#diff-97150b7a3ee0efd04dfd84457059fd3eR24) in the admin bundles controller too, and both the settings controller and bundles controller inherit from admin.

PR #501, opened 2016-06-14
adds Filter header to C4 table and centers CAT I and CAT III headers

MatthewPelland:


okeefm:
It looks like this commit is also present in #502, so I'm going to close this pull request. It should get pulled in when we merge in that one.

PR #504, opened 2016-06-15
Added load_library_functions to some setup/teardown methods

okeefm:
This should help to keep them in sync with the bundles that are installed in the testing environment, and maybe eliminate a few of the errors we're seeing.

### DIFF
--- a/test/controllers/admin/bundles_controller_test.rb
+++ b/test/controllers/admin/bundles_controller_test.rb
@@ -8,6 +8,10 @@ class Admin::BundlesControllerTest < ActionController::TestCase
     FileUtils.rm_rf(APP_CONFIG.bundle_file_path)
   end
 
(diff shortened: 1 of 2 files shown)


test/controllers/admin/bundles_controller_test.rb
```diff
@@ -8,6 +8,10 @@ class Admin::BundlesControllerTest < ActionController::TestCase
     FileUtils.rm_rf(APP_CONFIG.bundle_file_path)
   end
 
+  teardown do
```
ssayer: Why do this on teardown and not setup?

okeefm: So it gets run after the last tests, so the library functions aren't screwed up after this test suite finishes

PR #512, opened 2016-06-20
Added banner display variable to the admin dashboard

okeefm:


config/cypress.yml
```diff
@@ -9,7 +9,7 @@ default_bundle: "2.7.0"
 enable_logging: false
 
 #set to true to enable a banner at the top of every page with the "banner_message" text from below
-banner: false
+banner: <%= ENV["CYPRESS_BANNER"] || false %>
 banner_message: "This server is for demonstration purposes; data on it will be removed every Saturday at 11:59 PM Eastern Time"
```
tstrass: Is banner supposed to be separate from `banner_message`? Do we want `banner` configurable from the admin dashboard?

app/views/admin/_show_settings.html.erb
```diff
@@ -14,6 +14,15 @@
...
 
+<legend>Display Banner?</legend>
+<p class="configured-banner">
+  <% if banner %>
+    Yes
+  <% else %>
+    No
+  <% end %>
+</p>
+
```
tstrass: Lines 19-23 can be `<%= banner ? "Yes" : "No"%>

tstrass:
✅

PR #518, opened 2016-06-21
adds more information to user page

MatthewPelland:
now updated, alignment is less funky, account tab now displays user email truncated at 20 characters.
screen shot 2016-06-21 at 1 20 52 pm


ssayer:
* The alignment of the user data looks kinda funky
* listing the email is redundant since there's a field below
* Instead of "Account" in the nabber, we should display the user's email address (though we may have to truncate that display)

ssayer:
I'm good with this, @thecristen if you're ok with the design, please pull it in.

PR #532, opened 2016-06-22
Add ability for cypress.yml values to be pulled from environment.

rbclark:
This checks to see if a user has set different configuration variables in their environment before falling back to defaults directly defined in the cypress.yml.

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -24,12 +24,13 @@ class AdminController < ApplicationController
 
   def sub_yml_setting(key, val)
     yml_text = File.read("#{Rails.root}/config/cypress.yml")
+    sub_string = /#{key}:(.*?)\n/
     if val.is_a? String
-      yml_text.sub!(/^\s*#{key}: "(.*)"/, "#{key}: \"#{val}\"")
+      yml_text.sub!(sub_string, "#{key}: \"#{val}\"\n")
     elsif val.is_a? Symbol
-      yml_text.sub!(/^\s*#{key}: (.*)/, "#{key}: :#{val}")
+      yml_text.sub!(sub_string, "#{key}: :#{val}\n")
     else
-      yml_text.sub!(/^\s*#{key}: (.*)/, "#{key}: #{val}")
+      yml_text.sub!(sub_string, "#{key}: #{val}\n")
     end
     File.open("#{Rails.root}/config/cypress.yml", 'w') { |file| file.puts yml_text }
   end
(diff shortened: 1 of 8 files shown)


rbclark:
So this is going to actually take more consideration than I initially realized. The issue is that there are multiple calls to APP_CONFIG (for example https://github.com/projectcypress/cypress/blob/master/app/models/ability.rb#L10) which used to return true or false, but now returns as follows

    irb(main):001:0> APP_CONFIG.ignore_roles
    => "<%= if ENV['IGNORE_ROLES'].nil? then true else ENV['IGNORE_ROLES'] end %>"

which means that the data being pulled from cypress.yml is not actually being evaluated. One fix could be using `Settings[:ignore_roles]` however I do not know enough about the application to know the implications of such a change.

rbclark:
This did require tweaking the tests to set the Settings hash instead of the APP_CONFIG hash as well, however it seems that change did not break anything.

tstrass:
If you go into the admin dashboard and edit the settings, it completely blows away or doesn't replace any of the settings in cypress.yml now, because the regexs in [AdminController.sub_yml_settings](https://github.com/projectcypress/cypress/blob/update_cypress_variables/app/controllers/admin_controller.rb#L25) no longer match the values due to the ENV variable logic

tstrass:
Okay after giving this some thought here's what I think: Since we are now loading these values from environment variables, it no longer makes sense to overwrite values in cypress.yml. So we have two options:
1) We no longer persist any settings that are changed through the admin dashboard. We can update the values in memory, but they will be lost and reset to the environment variables on app restart.
2) We persist the new values by programmatically changing the environment variables. 

What do people think? @rbclark @okeefm @ssayer

okeefm:
I feel like we should still be able to persist values set through the admin dashboard, though I'm not sure what the best way to do so is.

rbclark:
It seems like it would be confusing to users if the settings kept switching and did not persist through reboots. I was looking to see if the Settings gem in use has a better way to handle this but it seems to always give ENV variables precedent which we do not want. What about the third option of having the program pull from the environment by default but subbing out the ruby code via regex. Looking at the regex it should be possible to just make it care less about how tightly it matches and just make it match up to the newline.

okeefm:
That would be an option. We're mostly using the environment variables to initialize settings, so once they're in the app, it would probably be okay to overwrite the yaml that pulls them in if local changes are made.

tstrass:
Okay that sounds good to me. I didn't realize the environment variables were only used for deployment.

tstrass:
✅

PR #557, opened 2016-07-01
Provider for Measure Test Records

jaebird123:
- added same provider for all records within a single measure test
- displayed provider information on test execution show page (only for measure tests)
- added integration and unit tests for this

screen shot 2016-07-01 at 9 03 20 am

### DIFF
--- a/app/models/measure_test.rb
+++ b/app/models/measure_test.rb
@@ -1,4 +1,6 @@
 class MeasureTest < ProductTest
+  belongs_to :provider, index: true
+
   validate :at_most_one_measure?
 
   def at_most_one_measure?
@@ -23,6 +25,11 @@ class MeasureTest < ProductTest
       C2Task.new(product_test: self).save!
       C3Cat3Task.new(product_test: self).save! if product.c3_test
     end
+  end
+
+  def generate_provider
+    self.provider = Provider.generate_provider(measure_type: measures.first.type)
+    save!
   end
 
   def archive_records
@@ -84,9 +91,5 @@ class MeasureTest < ProductTest
     else
       'unstarted'
     end
-  end
-
-  def provider
-    Provider.find(records.first.provider_performances.first.provider_id)
   end
 end
--- a/app/models/measure_test.rb
+++ b/app/models/measure_test.rb
@@ -85,4 +85,8 @@ class MeasureTest < ProductTest
       'unstarted'
     end
   end
+
+  def provider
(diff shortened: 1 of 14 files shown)


app/helpers/products_helper.rb
```diff
@@ -106,4 +106,12 @@ def with_c3_task(task)
     end
     [task]
   end
+
+  def provider_npi(provider)
```
ssayer: This code should probably be in the provider model

app/helpers/products_helper.rb
```diff
@@ -106,4 +106,12 @@ def with_c3_task(task)
...
+    provider.cda_identifiers.select { |cda| cda.root == '2.16.840.1.113883.4.6' }.first.extension # uses oid for NPI
+  end
+
+  def provider_tin(provider)
```
ssayer: Ditto

app/models/measure_test.rb
```diff
@@ -85,4 +85,8 @@ def cat3_status
       'unstarted'
     end
   end
+
+  def provider
+    Provider.find(records.first.provider_performances.first.provider_id)
```
ssayer: Can we directly assign the provider to the model, instead of having to trainwreck this?

app/helpers/products_helper.rb
jaebird123: so if the provider model is in health-data-standards, should we just extend the provider model in cypress?

ssayer: Actually, I just looked in HDS and it looks like it's in the Provider model already

https://github.com/projectcypress/health-data-standards/blob/master/lib/health-data-standards/models/provider.rb

app/models/measure_test.rb
```diff
@@ -85,4 +87,8 @@ def cat3_status
       'unstarted'
     end
   end
+
+  def provider
+    Provider.find(provider_id)
```
ssayer: Can't we' make this an active model relationship, otherwise it will make a db request every time you call .provider

lib/ext/provider.rb
```diff
@@ -1,4 +1,6 @@
 class Provider
+  has_many :product_tests
```
tstrass: Why not `:measure_tests`? Can they have filtering or checklist tests too?

jaebird123: good point, I'll change that

PR #562, opened 2016-07-05
Reduce AJAX Calls on Product Page

jaebird123:
- reduced the number of ajax requests on product show page by making single ajax request every 500ms from show.js.erb
- show.js.erb now only reloads partials if a task's product is not ready or a task has a test execution that is pending
- added integration test for AJAX reloading the product test links

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -106,4 +106,16 @@ module ProductsHelper
     end
     [task]
   end
+
+  # only yeilds tasks whose product test links need to be reloaded
+  def tasks_needing_reload(product)
+    tasks = []
+    product.product_tests.each do |test|
+      test.tasks.each do |task|
+        next unless %w(C1Task C2Task Cat1FilterTask Cat3FilterTask).include? task.class.to_s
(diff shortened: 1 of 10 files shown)


app/helpers/products_helper.rb
```diff
@@ -106,4 +106,16 @@ def with_c3_task(task)
     end
     [task]
   end
+
+  # only yeilds tasks whose product test links need to be reloaded
```
tstrass: yields @jaebird123

tstrass:
I added a product with a bunch of measures and it worked through all of them, but after it finished the last 6 were stuck on "building" after the requests stopped. @jaebird123

tstrass:
After adding a product, the bulk download progress no longer continuously updates. I think something is off with products/show.js.erb.

app/helpers/products_helper.rb
jaebird123: ah yes, i before e except after c and those other situations

jaebird123:
Ok, yeah I see the issue with not showing the progress of the bulk download. I'll fix that. I'll try to reproduce the measure tests stuck in `building` too and see if I can fix it

PR #567, opened 2016-07-06
Show warnings by file for passing executions

tstrass:
Previously, a passing execution did not display warnings by file.

**View:**

screen shot 2016-07-06 at 9 47 02 am

### DIFF
--- a/app/views/test_executions/_execution_results.html.erb
+++ b/app/views/test_executions/_execution_results.html.erb
@@ -6,8 +6,7 @@
 #
 %>
 
-<% case execution.status_with_sibling %>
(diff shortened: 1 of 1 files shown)


okeefm:
When I run this, this is what I get (not sure if I should have any warnings or not):

screen shot 2016-07-06 at 11 12 30 am

thecristen:
It looks like it doesn't check whether there are warnings - just displays the Warning text.
and @okeefm that's interesting that the test execution dropdown says (in progress)...

tstrass:
Good catch I'll fix that

PR #571, opened 2016-07-07
Bundle messages

schreiaj:
Add messages to Product and Test Execution pages if there is a `schematron_banner` key in the config for that bundle version.


ssayer:
I'm going to close this until the new alert is ready

PR #574, opened 2016-07-07
patient files should be html not xml

dczulada:


ssayer:
do we want to do this? if you're using QRDA for the patient import we would want a record of the QRDA looks like...

PR #575, opened 2016-07-07
prevents "Start" and "Upload" in measure test table from stacking.

MatthewPelland:


app/views/products/_measure_tests_table.html.erb
```diff
@@ -25,16 +25,16 @@
       </th>
     <% end %>
     <% if @product.c1_test %>
-      <th scope="col" class = 'no-wrap pointer-on-hover'>
+      <th scope="col" style = 'min-width: 100px' class = 'no-wrap pointer-on-hover'>
```
thecristen: let's use a CSS class for the `min-width: 100px` instead of a style attribute.

app/views/products/_measure_tests_table.html.erb
```diff
@@ -25,16 +25,16 @@
...
       </th>
     <% end %>
     <% if @product.c2_test %>
-      <th scope="col" class = 'no-wrap pointer-on-hover'>
+      <th scope="col" style = 'min-width: 100px' class = 'no-wrap pointer-on-hover'>
         <%= c2_label %>
       </th>
     <% end %>
-    <th scope="col" class="no-wrap">
+    <th scope="col" style = 'width:10%' class="no-wrap">
```
thecristen: Maybe using the `col-sm-2` class might work here instead?

PR #581, opened 2016-07-11
value set should not be included if there is a negation, and the nega…

dczulada:
value set should not be included if there is a negation, and the negation doesn't match


dczulada:
Was

screen shot 2016-07-11 at 12 10 43 pm

Should be 

screen shot 2016-07-11 at 12 10 54 pm

PR #591, opened 2016-07-14
Removes .first from test suite where possible

MatthewPelland:
In some cases .first was used with a set of randomized elements, in which case find() or find_by wouldn't work as each element has unknown characteristics
I ran all the tests a few times, but they had been breaking some times but not others, if that happens it's probably my fault and I can fix it

### DIFF
--- a/features/step_definitions/product.rb
+++ b/features/step_definitions/product.rb
@@ -178,13 +178,13 @@ When(/^all product tests have a state of ready$/) do
 end
 
 When(/^all product tests do not have a state of ready$/) do
-  pt = ProductTest.first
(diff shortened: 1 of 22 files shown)


features/step_definitions/product.rb
```diff
@@ -178,13 +178,14 @@ def build_product
 end
 
 When(/^all product tests do not have a state of ready$/) do
-  pt = ProductTest.first
+  pt = ProductTest.find_by(cms_id: 'CMS155v1')
+  byebug
```
ssayer: This needs to be removed

PR #592, opened 2016-07-14
Execution errors

tstrass:
# Summary:

If an exception was thrown internally during a test execution, the execution would permanently and inaccurately show as "In progress". This adds the status of "Errored" to test executions. It follows that this adds the status of "Errored" to products as well (if any tests are failing, the "Failing" status takes precedence over "Errored").

# Views:

### vendors#index
screen shot 2016-07-14 at 12 21 04 pm

### vendors#show
screen shot 2016-07-14 at 12 21 34 pm

### product#show
screen shot 2016-07-14 at 12 13 04 pm
screen shot 2016-07-14 at 12 21 59 pm

### test_executions#show
screen shot 2016-07-14 at 12 17 44 pm

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -47,13 +47,15 @@ module ProductsHelper
 
     case test_status
     when 'passing'
-      return 5
+      return 6
     when 'failing'
+      return 5
+    when 'errored'
       return 4
     when 'incomplete'
       return 3
     else
-      return 6
+      return 7
     end
   end
 
@@ -86,6 +88,7 @@ module ProductsHelper
     return tasks.first.status if tasks.count == 0
     return 'passing' if tasks.all?(&:passing?)
     return 'failing' if tasks.any?(&:failing?)
+    return 'errored' if tasks.any?(&:errored?)
     return 'incomplete' if tasks.any?(&:incomplete?)
     'unstarted'
   end
--- a/app/helpers/vendors_helper.rb
+++ b/app/helpers/vendors_helper.rb
@@ -38,8 +38,8 @@ module VendorsHelper
 
   def c3_status_values(product)
     h = {}
-    cat1_status_values = product.c1_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat1Task') : [0, 0, 0, 0]
-    cat3_status_values = product.c2_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat3Task') : [0, 0, 0, 0]
+    cat1_status_values = product.c1_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat1Task') : [0, 0, 0, 0, 0]
+    cat3_status_values = product.c2_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat3Task') : [0, 0, 0, 0, 0]
     h['QRDA Category I'] = Hash[%w(passing failing errored not_started total).zip(cat1_status_values)]
     h['QRDA Category III'] = Hash[%w(passing failing errored not_started total).zip(cat3_status_values)]
     h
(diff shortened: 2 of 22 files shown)


app/helpers/products_helper.rb
```diff
@@ -47,13 +47,15 @@ def set_sorting(test, test_status)
 
     case test_status
     when 'passing'
-      return 5
+      return 6
```
okeefm: Yay magic numbers. I understand these were here before, but if we ever go through and clean up code, we should replace these with constants or something.

app/helpers/vendors_helper.rb
```diff
@@ -19,56 +19,59 @@ def get_product_status_values(product)
...
 
   def c3_status_values(product)
     h = {}
-    cat1_status_values = product.c1_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat1Task') : [0, 0, 0, 0]
-    cat3_status_values = product.c2_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat3Task') : [0, 0, 0, 0]
-    h['QRDA Category I'] = Hash[%w(passing failing not_started total).zip(cat1_status_values)]
-    h['QRDA Category III'] = Hash[%w(passing failing not_started total).zip(cat3_status_values)]
+    cat1_status_values = product.c1_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat1Task') : [0, 0, 0, 0, 0]
+    cat3_status_values = product.c2_test ? product_test_statuses(product.product_tests.measure_tests, 'C3Cat3Task') : [0, 0, 0, 0, 0]
+    h['QRDA Category I'] = Hash[%w(passing failing errored not_started total).zip(cat1_status_values)]
```
dehall: this word array is repeated all over the place, should probably be a constant

app/helpers/products_helper.rb
thecristen: These numbers are set on a cell to dictate how DataTables sorts a given column... maybe I'm just lacking imagination but how might replacing these with constants work?

app/views/test_executions/_execution_results.html.erb
```diff
@@ -131,4 +131,19 @@
...
+  <% [execution, execution.sibling_execution].each do |ex| %>
+    <% ex_type = ex.task._type[0, 2] %>
+    <% if ex.errored? %>
+      <p class="lead row bg-warning execution-status"><i class="fa fa-fw fa-exclamation-circle text-warning" aria-hidden="true"></i> <%= ex_type %> Execution: An internal error occured</p>
+    <% elsif ex.passing? %>
+      <p class="lead row bg-success execution-status"><i class="fa fa-fw fa-check-circle text-success"></i> <%= ex_type %> Execution passed</p>
+    <% elsif ex.failing? %>
+      <p class="lead row bg-danger execution-status">
+        <i class="fa fa-fw fa-times-circle text-danger" aria-hidden="true"></i>
+        <%= execution_failure_message(ex) %>
```
thecristen: Might want to add `ex_type` to the failure message as well as the errored and passing ones.

app/models/test_execution.rb
```diff
@@ -46,6 +46,8 @@ def validate_artifact(validators, artifact, options = {})
                              :msg_type => :error, :validator_type => :result_validation, :validator => :smoking_gun)
     end
     execution_errors.only_errors.count > 0 ? fail : pass
+  rescue
+    error
```
thecristen: Might having a _method_ called `error` be confusing for other devs? @okeefm @dehall ?

okeefm: I thought about this too. Maybe name the method `errored` instead?

app/helpers/products_helper.rb
okeefm: Instead of returning `5`, return something like `TABLE_FIRST`, which is set to 0. I guess it makes a little more sense when I know that it's a sort order thing. Maybe just a comment to that effect would be sufficient.

tstrass: I think it might make even more sense to make an array of the possible statuses and then set the sort order based off indexes in that array. Easier to change around the order or add a status that way

app/helpers/vendors_helper.rb
tstrass: yeah maybe that should go in a separate refactor PR though

app/views/test_executions/_execution_results.html.erb
tstrass: I intentionally left that out because I thought it would be weird to have the entire super long XML viewer that comes with a failure and then after all of that another execution result message. It just occurred to me that we'd have this same problem though with a test that passes with warnings. I'll move some things around

tstrass:
Updated view of execution results, in the case of one error and one failure:
screen shot 2016-07-14 at 3 40 03 pm

tstrass:
Note this branch was just rebased to accommodate #587

PR #595, opened 2016-07-14
fixes admin breadcrumbs and makes panel buttons consistent

MatthewPelland:
Admin breadcrumbs were doubled, and panel buttons didn't have consistent wording.


ssayer:
This just needs rubocop fixes

PR #598, opened 2016-07-18
fix c1 and bulk download links

ssayer:
This is so @jaebird123 can review these changes while I'm in a meeting

### DIFF
--- a/app/views/application/_checklist_execution_results.html.erb
+++ b/app/views/application/_checklist_execution_results.html.erb
@@ -24,7 +24,7 @@
     <tr>
       <td><%= link_to 'View Results', new_task_test_execution_path(execution.task) %></td>
       <% # for both c1 and c3 executions (if they exist) %>
-      <% [execution, c3_execution].reject { |exec| exec.nil? }.each do |exec| %>
+      <% [execution, c3_execution].compact.each do |exec| %>
         <td>
           <% case exec.status %>
           <% when 'passing' %>
--- a/app/views/application/_checklist_execution_results.html.erb
+++ b/app/views/application/_checklist_execution_results.html.erb
@@ -24,7 +24,7 @@
     <tr>
       <td><%= link_to 'View Results', new_task_test_execution_path(execution.task) %></td>
       <% # for both c1 and c3 executions (if they exist) %>
-      <% [execution, c3_execution].reject { |exec| exec == false }.each do |exec| %>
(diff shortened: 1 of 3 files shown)


app/views/application/_checklist_execution_results.html.erb
```diff
@@ -24,7 +24,7 @@
     <tr>
       <td><%= link_to 'View Results', new_task_test_execution_path(execution.task) %></td>
       <% # for both c1 and c3 executions (if they exist) %>
-      <% [execution, c3_execution].reject { |exec| exec == false }.each do |exec| %>
+      <% [execution, c3_execution].reject { |exec| exec.nil? }.each do |exec| %>
```
jaebird123: cool, this change makes sense. I didn't realize the `sibling_execution()` function on a `test_execution` returned `nil` and not `false`

ssayer: In this case it does, in other cases it might still return false. The bug I'm trying to fix happens in C1 only, so it's possible that C1/C2 tests return false.

app/views/products/_bulk_download.html.erb
```diff
@@ -10,7 +10,7 @@
 <% num_measure_tests_ready = product.product_tests.measure_tests.where(state: :ready).count %>
 <% if num_measure_tests_ready == num_measure_tests %>
   <p>This download contains a folder for each measure selected for this product. Inside these folders are XML documents for each patient associated with that measure.</p>
-  <%= form_for product, url: { controller: 'products', action: 'patients' }, :html => { :method => 'GET' } do |f| %>
+  <%= form_for download_full_test_deck_product_path(product), :html => { :method => 'GET' } do |f| %>
```
jaebird123: I'm not sure about this one. I think we removed the `download_full_test_deck` route in place of a `patients` route in the product controller. maybe this should be `patients_vendor_product_path(product.vendor, product)`?

app/views/application/_checklist_execution_results.html.erb
jaebird123: I think they shou

app/views/products/_bulk_download.html.erb
ssayer: The route is still in there. If there's a more appropriate one to use, we can use that, but we shouldn't be using url: controller/action unless we have a really good reason.

PR #608, opened 2016-07-19
Makes manual entry fields big enough to show codes

MatthewPelland:
Splits Manual Entry rows into two rows, allowing input boxes to be large enough to fit codes
screen shot 2016-07-19 at 3 27 18 pm


ssayer:
@MatthewPelland screenshot plz

PR #622, opened 2016-07-21
Point packer to RC1, dynamically set version (as allowed in new chef recipe).

rbclark:
This PR makes use of the new option in our chef recipe to regenerate the secret_key_base on first boot in order to remove that step from the instructions. Also it updates to rc1.


rbclark:
Not sure why this is failing, it shouldn't touch anything that the tests even run.

okeefm:
It's because our test suite is horribly glitchy on Travis. I'll re-run it.

PR #631, opened 2016-07-25
if there isn't a patient for the IPP of a submeasure, grab one from s…

dczulada:
…ub population a


dczulada:
Matt, were you going to post an update here.

MatthewPelland:
it's all set, the problem's been fixed

PR #632, opened 2016-07-25
fixes filename anchors in case they contain spaces

dehall:


app/helpers/test_executions_results_helper.rb
```diff
@@ -78,6 +78,11 @@ def compare_error_locations_across_files(a, b)
     a.location <=> b.location
   end
 
+  def filename_slug(execution, file_name, use_selector = false)
+    selector = use_selector ? '#' : ''
```
thecristen: I get what `use_selector` does, but part of me wants to leave it out of this function and have the developer type the '#' before anchors when writing markup...

PR #634, opened 2016-07-25
allows symbols (e.g. ) in file names for QRDA uploads

MatthewPelland:
some symbols caused problems with ASCII-8BIT versus UTF-8 in string comparisons.
previously these files just didn't show up in the list, even when there were errors


ssayer:
rubocop

PR #635, opened 2016-07-25
sorts MPL and patient list for tests alphabetically

MatthewPelland:


ssayer:
@MatthewPelland rubocop

PR #642, opened 2016-07-26
hds bump

ssayer:
Bump HDS for latest code.


dczulada:
One more HDS update needed.  The latest HL7 schematron was posted 11 days ago.

PR #644, opened 2016-07-26
finish making all the times consistent and in local time

thecristen:


ssayer:
@thecristen There was bug that made it's way into master and is causing builds to break, can your rebase your PR?

thecristen:
@ssayer done

PR #657, opened 2016-07-27
Both collapsible navbar and delete window work now

MatthewPelland:


app/assets/javascripts/application.js
```diff
@@ -27,8 +27,6 @@
 //= require assets_framework/breadcrumb
 //= require jquery_nested_form
 //= require jasny-bootstrap.min
-//= require bootstrap.min
-//= require local_time
```
thecristen: Why'd you delete `local_time`?

app/assets/javascripts/application.js
```diff
@@ -3,7 +3,6 @@
 //= require turbolinks
 //= require jquery
 //= require jquery_ujs
-//= require jquery.remotipart
```
thecristen: Why is this leaving us?

MatthewPelland:
I have no clue where those went, I even ran git diff locally and there was only the one addition...

PR #663, opened 2016-07-28
fixes bundle issues when downloading c1 manual patients

MatthewPelland:


dczulada:
There is still a Bundle.default on line 44

PR #665, opened 2016-07-28
one more Bundle.default in print_criteria

dczulada:


MatthewPelland:
sorry, that line should actually just be deleted, it's no longer needed after my last pr

MatthewPelland:
good to go now

PR #680, opened 2016-08-08
creates downloadable zip file of all patients

MatthewPelland:


satyakirankumar:
Provide the Summary of the Results also.

ssayer:
@satyakirankumar There isn't really a good way for us to do that, as there is no standard for displaying this information.

PR #688, opened 2016-08-15
Vendor and Product deletion Optimization

rbclark:
This cleans up the destroy commands to run many drop commands all at once in order instead of one by one in order to save from an expensive activerecord callback to every item.

**Results:**

    # Old product deletion code:

    Measure Mode: wall_time
    Thread ID: 70211009151460
    Fiber ID: 70211092997080
    Total: 1.391805
    Sort by: self_time

    # New product deletion code:

    Measure Mode: wall_time
    Thread ID: 70102058131940
    Fiber ID: 70102131343280
    Total: 0.180934
    Sort by: self_time

    # Old vendor deletion code:

    Measure Mode: wall_time
    Thread ID: 70115970128860
    Fiber ID: 70116032985480
    Total: 403.965664
    Sort by: self_time

    # New vendor deletion code:

    Measure Mode: wall_time
    Thread ID: 70106575397360
    Fiber ID: 70106659652940
    Total: 1.284944
    Sort by: self_time 

I can commit the code used to generate the test results if desired.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -101,6 +101,14 @@ class Product
     end
   end
 
+  # This method does nothing more than attempt to cleanup a lot of data instead of making rails do it,
+  # since rails is really bad at cleaning up quickly.
+  def destroy
+    ProductTest.destroy_by_ids(product_test_ids)
+
+    super
+  end
+
   # - - - - - - - - - #
   #   P R I V A T E   #
   # - - - - - - - - - #
--- a/app/models/vendor.rb
+++ b/app/models/vendor.rb
@@ -34,6 +34,20 @@ class Vendor
     end
   end
 
+  # This method does nothing more than attempt to cleanup a lot of data instead of making rails do it,
+  # since rails is really bad at cleaning up quickly. Note that this bypasses the dependent => destroy
+  # calls on product, product tests, etc, all the way down and does them manually. This means that
+  # if any of those structures change then this code will need to be updated accordingly.
+  def destroy
+    product_tests = ProductTest.where(:product_id.in => product_ids)
+    product_test_ids = product_tests.pluck(:_id)
+    ProductTest.destroy_by_ids(product_test_ids)
+
+    Product.in(id: product_ids).delete
+
+    super
+  end
+
   def status
     Rails.cache.fetch("#{cache_key}/status") do
       total = products.count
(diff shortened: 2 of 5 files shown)


rbclark:
So there are currently 2 things messing with this PR. First up there is apparently a vulnerability in our rails version and we need to bump. Second I've apparently managed to push the line length of product.rb over 100 lines and now it is complaining that it is too long. First one is an easy fix but I could use suggestions on how to clean up the second one.

rbclark:
Okay I believe everything is fixed with this but the class length. I am not exactly sure how I would go about making this much shorter. The two destroy functions are different enough that I am not sure it would be beneficial to try to merge them and put them in some sort of helper and it may be confusing to later developers if the destroy function is outside of the model.

app/models/product.rb
```diff
@@ -101,6 +101,24 @@ def add_checklist_test
...
+  def destroy
+    tasks = Task.where(:product_test_id.in => product_test_ids)
+    records = Record.where(:test_id.in => product_test_ids)
+    record_ids = records.pluck(:_id)
+    calculation_results = HealthDataStandards::CQM::PatientCache.where(:'value.patient_id'.in => record_ids)
+
+    calculation_results.delete
+    records.delete
+    tasks.delete
+
```
okeefm: `Task`s have many `TestExecution`s, so we should probably delete them as well, when we delete both the vendor and the product

rbclark:
@okeefm I have updated this to remove test executions and Artifacts along with their associated files as well as suggested.

app/models/vendor.rb
```diff
@@ -34,6 +34,20 @@ def self.accessible_by(user)
...
+  # calls on product, product tests, etc, all the way down and does them manually. This means that
+  # if any of those structures change then this code will need to be updated accordingly.
+  def destroy
+    product_tests = ProductTest.where(:product_id.in => product_ids)
+    product_test_ids = product_tests.pluck(:_id)
+    ProductTest.destroy_by_ids(product_test_ids)
+
+    Product.in(id: product_ids).delete
+
+    super
```
okeefm: I assume if everything else is already deleted above, this `super` call doesn't slow everything down with callbacks (because there's nothing to callback on?)

rbclark: Correct, the call to super doesn't slow anything down and it still calls the before_destroy and after_destroy methods for the vendor model, and will delete any dependent associations that were missed somehow.

PR #691, opened 2016-08-16
errors for data criteria outside the measure definition are now warnings

dczulada:


dczulada:
We still don't know if we're doing this yet...

okeefm:
Any chance of getting a test for this?

PR #698, opened 2016-08-23
show measure calculation unless the list is for a product test, Cypre…

dczulada:
…ss is in ATL Mode and the current user is not an ATL

### DIFF
--- a/app/helpers/records_helper.rb
+++ b/app/helpers/records_helper.rb
@@ -64,7 +64,7 @@ module RecordsHelper
   end
 
   def hide_patient_calculation?
-    # Hide measure calculation if it is a product test, Cypress is in ATL Mode and the current user is not an ATL or admin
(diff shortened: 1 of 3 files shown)


app/helpers/records_helper.rb
```diff
@@ -62,4 +62,9 @@ def display_field(field)
   def coverage_for_measure(measure_id, sub_id, bundle_id)
     QueryCache.where(measure_id: measure_id, sub_id: sub_id, bundle_id: bundle_id, test_id: nil).first['bonnie_coverage']
   end
+
+  def hide_patient_calculation?
+    # Hide measure calculation if it is a product test, Cypress is in ATL Mode and the current user is not an ATL or admin
+    @product_test && ApplicationController.helpers.mode_atl? && (!current_user.user_role?('admin') && !current_user.user_role?('atl'))
```
ssayer: product test should be explicitly passed, it's not good to rely on an instance variable outside the method. Also, is there a cleaner way to call/access ApplicationController.helpers.mode_atl?.

PR #701, opened 2016-08-29
a user should be able to view other submeasures calculation results

dczulada:
When a measure has sub measuers
screen shot 2016-08-29 at 3 19 11 pm

When a measure does not
screen shot 2016-08-29 at 3 21 20 pm


app/views/records/index.html.erb
```diff
@@ -1,8 +1,22 @@
 <% if @task && @product_test %>
   <h1>Patient List</h1>
-  <%= button_to html_patients_product_test_path(@product_test), :method => :get, :class => "btn btn-default pull-right" do %>
-    <i class="fa fa-fw fa-download" aria-hidden="true"></i> Download HTML Patients
-  <% end %>
+  <div>
```
thecristen: I'd add the `pull-right` class to this div and remove it from the button and dropdown (and you'll have to switch the button and dropdown so they appear in the correct order). Otherwise looks great!

dczulada:
screen shot 2016-08-30 at 12 16 13 pm

thecristen:
Was totally expecting these to end up side by side... 👎

app/views/records/index.html.erb
```diff
@@ -1,21 +1,23 @@
...
+          </button>
+          <ul class="dropdown-menu">
+            <% submeasures = HealthDataStandards::CQM::Measure.where(hqmf_id: @measure.hqmf_id).sort_by(&:sub_id) %>
+            <% submeasures.each do |submeasure| %>
+              <li><%= link_to "#{submeasure.display_name} - #{submeasure.subtitle}", { controller: 'records', task_id: @task.id, sub_id: submeasure.sub_id}, method: :get %></li>
+            <% end %>
+          </ul>
+        </div>
+      <% end %>
+    </div>
```
thecristen: Ideally the two buttons would be side by side.
What if you switch the button position?

```erb
<% if @measure && @measure.sub_id %>
         <div class="btn-group">
           <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown">Sub measures<span class="caret"></span>
           </button>
           <ul class="dropdown-menu">
             <% submeasures = HealthDataStandards::CQM::Measure.where(hqmf_id: @measure.hqmf_id).sort_by(&:sub_id) %>
             <% submeasures.each do |submeasure| %>
               <li><%= link_to "#{submeasure.display_name} - #{submeasure.subtitle}", { controller: 'records', task_id: @task.id, sub_id: submeasure.sub_id}, method: :get %></li>
             <% end %>
           </ul>
         </div>
<% end %>
<%= button_to html_patients_product_test_path(@product_test), :method => :get, :class => "btn btn-default" do %>
         <i class="fa fa-fw fa-download" aria-hidden="true"></i> Download HTML Patients
<% end %>
```

Or maybe they should both be wrapped in a `.btn-group`? I can probably look at this more tomorrow if you don't figure it out by then

dczulada: That would be great...i'm banging my head trying to figure out this one.

dczulada:
Side by side!

screen shot 2016-09-01 at 2 02 26 pm

PR #706, opened 2016-09-01
Optimize requests on the vendor product page

rbclark:
Instead of pulling each row of the different tables one by one, pull them all at once and only pull one by one when individual tests are uploaded.

Results before when creating a new product and staying on the page for the duration of the processing:
before

Results after:
after

Tests for this may be failing at the moment, I just rebased against master however it seems master may be failing as well.


thecristen:
Does this still allow the table to be updated as individual measure tests are being built?

rbclark:
@thecristen Yes, the UI still updates nearly the same, a lot of tasks are just done in bulk and a little less often.

PR #713, opened 2016-09-14
Cypress 3.0.1 patches

ssayer:


app/views/application/_header_product.html.erb
```diff
@@ -4,7 +4,7 @@
       <%= button_to edit_vendor_product_path(@product.vendor, @product), :method => :get, :class => "btn btn-default" do %>
         <i class="fa fa-fw fa-wrench" aria-hidden="true"></i> Edit Product
       <% end %>
-      <% if mode_atl?%>
+      <% if current_user.user_role?([:atl, :admin]) %>
```
okeefm: Is this meant to not fire in demo mode, then? If the user isn't an ATL or an admin, this will only be true in precert mode, not in demo mode.

config/cypress.yml
```diff
@@ -3,7 +3,7 @@ effective_date:
   month: 12
   day: 31
 
-version: 3.0.0
+version: 3.0.1
```
okeefm: We should also bump the version number in the build scripts, so we can build new VMs, etc.

app/views/application/_header_product.html.erb
okeefm: I'm also getting a failure in the `user can download report in ATL mode` cucumber task; both on my machine and on Travis

ssayer: It wouldn't have fired in Demo mode before, since it would only fire in ATL mode. We've been trying to move away from using the modes for permissions and use the user_role? method so that it always works in precert mode. I suppose we could add demo mode here though...

config/cypress.yml
ssayer: @rbclark?

rbclark: Already completed, https://github.com/projectcypress/cypress/pull/696

app/views/application/_header_product.html.erb
okeefm: I'll merge it; we can make the decision later about demo mode, given that we're really the only users of a demo mode server.

PR #721, opened 2016-09-21
Don't include prev and next buttons in manual test results page

tstrass:
This should fix [CYPRESS-724](https://oncprojectracking.healthit.gov/support/browse/CYPRESS-769) by removing the buttons to cycle through other product tests on the checklist results page. Doing that didn't make sense, because a product only ever has one checklist test.

### DIFF
--- a/app/views/test_executions/show.html.erb
+++ b/app/views/test_executions/show.html.erb
@@ -10,14 +10,16 @@
   <% cat3_task = @task.product_test.tasks.cat3_filter_task %>
 <% end %>
 <%= render partial: 'application/certification_bar', locals: { product: @product_test.product, active_certs: current_certifications(@task._type, @product_test.product.c3_test) } %>
-<div class="panel clearfix">
-  <%= button_to new_task_test_execution_path(iterate_task(@task, 'prev').id), :method => :get, :class => "btn btn-default pull-left" do %>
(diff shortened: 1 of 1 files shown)


app/views/test_executions/show.html.erb
```diff
@@ -10,14 +10,16 @@
...
 <%= render partial: 'application/certification_bar', locals: { product: @product_test.product, active_certs: current_certifications(@task._type, @product_test.product.c3_test) } %>
-<div class="panel clearfix">
-  <%= button_to new_task_test_execution_path(iterate_task(@task, 'prev').id), :method => :get, :class => "btn btn-default pull-left" do %>
-    <i class="fa fa-fw fa-step-backward" aria-hidden="true"></i> Previous Test: <%= iterate_task(@task, 'prev').product_test.cms_id %>
-  <% end %>
-  <%= button_to new_task_test_execution_path(iterate_task(@task, 'next').id), :method => :get, :class => "btn btn-default pull-right" do %>
-    Next Test: <%= iterate_task(@task, 'next').product_test.cms_id %> <i class="fa fa-fw fa-step-forward" aria-hidden="true"></i>
-  <% end %>
-</div>
+<% unless @task._type == 'C1ManualTask' || @task._type == 'C3ManualTask' %>
```
ssayer: So I'm not going to block this PR just for this, but we really should create some kind of helper method or something for tasks so we don't have to do == 'some string' everywhere. I'll add a chore to pivotal for it.

PR #726, opened 2016-09-27
strat should be in its own row

dczulada:


tstrass:
@dczulada would you mind just posting a screenshot? I tried to make an example but I don't know how to cause reporting errors.

dczulada:
screen shot 2016-09-28 at 9 24 39 am

tstrass:
I know this isn't something new with this PR, but is there a reason we have reporting errors grouped into different tables, and if so should there be some kind of indication of what the groupings mean?

dczulada:
Yes they should be grouped

screen shot 2016-09-28 at 11 26 50 am

screen shot 2016-09-28 at 11 27 20 am

PR #730, opened 2016-09-29
separate user_role?([atl, admin]) into 2 calls

dczulada:


tstrass:
It looks like the Rolify has a method for this purpose https://github.com/RolifyCommunity/rolify/wiki/Usage#multiple-role-checking @dczulada

tstrass:
Oh yes but we need our `user_role` function to account for if `ignore_roles` is on, so nevermind. Unless you want to make an `user_any_role` method

PR #741, opened 2016-10-05
Removed manual comparison of task type from multiple points in the code

rbclark:
Added helper methods in order to add the ability to call (for example) `task.C1Task?` instead of having to call `task._type == 'C1Task'`. Dynamically generate methods in order to reduce code duplication.

The biggest objection I can see to this PR is the dynamic method generation. It makes the code a lot more concise and makes it a lot easier to add additional test states or states if ever needed, however it adds a little more of a learning curve to the code.

### DIFF
--- a/app/controllers/test_executions_controller.rb
+++ b/app/controllers/test_executions_controller.rb
@@ -20,7 +20,7 @@ class TestExecutionsController < ApplicationController
     authorize! :execute_task, @task.product_test.product.vendor
     @test_execution = @task.execute(results_params)
     respond_with(@test_execution) do |f|
-      if @task._type == 'C1ManualTask'
+      if @task.is_a? C1ManualTask
         f.html { redirect_to product_checklist_test_path(@task.product_test.product, @task.product_test) }
       else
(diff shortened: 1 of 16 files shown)


tstrass:
Why not get rid of the string comparison all together just use reflection (`is_a?`) Instead of the metaprogramming for test and task types?
Examples where that's already used for [tasks](https://github.com/projectcypress/cypress/blob/cleanupLogicChecks/app/views/test_executions/_expected_results.html.erb#L1) and [tests](https://github.com/projectcypress/cypress/blob/cleanupLogicChecks/app/controllers/application_controller.rb#L94).

I do like other the metaprogramming you added [here](https://github.com/projectcypress/cypress/pull/741/files#diff-6d99daed20a86e22fcc84e9810fc3e23R108) and [here](https://github.com/projectcypress/cypress/pull/741/files#diff-c95c9451cf097bdd58f73d0f9b114f97R38) though.

tstrass:
But if you stick with what you have for test and task types, we probably oughta change the the places where reflection is used to use your new mechanism as well.

rbclark:
I like your method of using reflection better, I'll update my stuff accordingly.

PR #742, opened 2016-10-06
A few small optimization tweaks

rbclark:
Removed duplicated counting of the number of products in a vendor. Use vendor_id instead of vendor when generating a link to the vendor product path since calling product.vendor generates another database query (for the complete vendor) but vendor_id doesn't.


rbclark:
Yep I overlooked that one by mistake, I've moved that over accordingly.

PR #743, opened 2016-10-07
Added port range validation to smtp settings

rbclark:
I was initially planning to add validation for all the fields however any additional validation has the potential to block legitimate settings. For example filtering the SMTP server name to only hostnames would mean that things such as localhost or an IP would be considered invalid, even though they could be perfectly valid settings. The case for the mailer username is similar since it could potentially be a full email or a single word and either would be valid.

### DIFF
--- a/app/views/admin/settings/edit.html.erb
+++ b/app/views/admin/settings/edit.html.erb
@@ -15,7 +15,7 @@
         <legend>Email Settings</legend>
         <div class="row">
           <%= f.text_field :mailer_address, wrapper: { class: "col-sm-10" },  label: "SMTP Server", value: smtp_settings.address %>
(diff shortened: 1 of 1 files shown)


app/views/admin/settings/edit.html.erb
```diff
@@ -15,7 +15,7 @@
         <legend>Email Settings</legend>
         <div class="row">
           <%= f.text_field :mailer_address, wrapper: { class: "col-sm-10" },  label: "SMTP Server", value: smtp_settings.address %>
-          <%= f.number_field :mailer_port, wrapper: { class: "col-sm-2" }, label: "Port", maxLength: 5, value: smtp_settings.port %>
+          <%= f.number_field :mailer_port, wrapper: { class: "col-sm-2" }, label: "Port", maxLength: 5, value: smtp_settings.port, min: 1, max: 65535 %>
```
okeefm: Do we still need/want the `maxLength` validation, if we have a min/max validation?

PR #744, opened 2016-10-17
update hds

dczulada:


okeefm:
I'm getting the same failure that CI is getting on my local machine when I run `bundle exec rake cucumber`.

Gemfile.lock
```diff
@@ -254,7 +254,7 @@ GEM
     mocha (1.1.0)
       metaclass (~> 0.0.1)
     modernizr-rails (2.7.1)
-    mongo (2.2.7)
+    mongo (2.3.0)
```
okeefm: This is causing the CI failures. We've had this problem before; we should decide whether we want to fix it, or pin `mongo` to `< 2.3.0`.

okeefm:
If you open the Cypress page from the failing Cucumber test, it contains the following text: `Cypress encountered an internal error while building this measure. Please contact the Cypress team at https://github.com/projectcypress/cypress/issues, and include the text below: undefined method `+' for nil:NilClass`. The given HQMF ID is `40280381-4BE2-53B3-014C-0F589C1A1C39`

features/support/env.rb
```diff
@@ -139,11 +139,16 @@ def value_or_bson(v)
...
   json.each_pair do |k, v|
-    if v.is_a? Hash
+    if v.is_a? Array
+      v_array = []
+      v.each do |av|
+        v_array << value_or_bson(av)
+      end
+      json[k] = v_array
+    elsif v.is_a? Hash
       json[k] = value_or_bson(v)
```
okeefm: can replace the Array block with `json[k] = v.map { |av| value_or_bson(av) }`

PR #745, opened 2016-10-17
Fail in a more clean manner when the upgrade script runs into an error  involving cypress.yml

rbclark:
When merging, if the changes between upstream and local cypress.yml do not merge cleanly it currently leaves the app in a broken state. This fixes it by reverting to the upstream version (which will reset any user settings in the file.)

I can pretty easily make this revert to the users version instead of the upstream version (thus preserving their settings but not pulling in changes to the cypress.yml file). I am not sure which is more desirable in this case.

### DIFF
--- a/contrib/upgrade_cypress_30.sh
+++ b/contrib/upgrade_cypress_30.sh
@@ -30,6 +30,16 @@ function pull_git_tag() {
     # pull in case we're on a branch
     sudo -u cypress git pull
     sudo -u cypress git stash pop
+    rc=$?
+    if [[ $rc != 0 ]]; then
+      sudo -u cypress cp config/cypress.yml config/cypress.yml.old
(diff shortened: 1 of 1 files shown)


contrib/upgrade_cypress_30.sh
```diff
@@ -30,6 +30,16 @@ function pull_git_tag() {
     # pull in case we're on a branch
     sudo -u cypress git pull
     sudo -u cypress git stash pop
+    rc=$?
+    if [[ $rc != 0 ]]; then
+      sudo -u cypress cp config/cypress.yml config/cypress.yml.old
+      echo "WARNING: Unable to merge cypress config cleanly, you will need to manually reset your cypress settings."
```
okeefm: Might make sense to point them directly at the file they need to change, in addition to where they can find the old version

PR #746, opened 2016-10-20
code/valueset check in manual entry form should also take bundle into…

dczulada:
… account

### DIFF
--- a/app/models/checklist_source_data_criteria.rb
+++ b/app/models/checklist_source_data_criteria.rb
@@ -54,7 +54,7 @@ class ChecklistSourceDataCriteria
                  else
                    [criteria.negation_code_list_id]
                  end
-      self.attribute_complete = code_in_valuesets(valueset, attribute_code)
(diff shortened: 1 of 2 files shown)


app/models/checklist_source_data_criteria.rb
```diff
@@ -117,7 +117,7 @@ def compare_field_values(data_criteria, criteria)
   end
 
   # searches an array of valuesets for a code
-  def code_in_valuesets(valuesets, input_code)
-    !HealthDataStandards::SVS::ValueSet.where('concepts.code' => input_code).in(oid: valuesets).empty?
+  def code_in_valuesets(valuesets, input_code, bundle_id)
+    !HealthDataStandards::SVS::ValueSet.where('concepts.code' => input_code).in(oid: valuesets, bundle_id: bundle_id).empty?
```
okeefm: Should the `bundle_id: bundle_id` be in the `where` statement, rather than the `in` statement?

PR #752, opened 2016-11-01
MPL download updates

okeefm:
* Created an `MplDownloadCreateJob` to create the MPL download zip on demand (either when bundle is uploaded, or if it doesn't exist, when the button is clicked)
* Added a task to the `BundleUploadJob` to call the `MplDownloadCreateJob`
* Separated out the downloading of multiple bundles, so each bundle now has its own "Download" button
* Improved the bundle download create task, by creating it on disk first then zipping it up for efficiency

### DIFF
--- a/app/views/records/index.html.erb
+++ b/app/views/records/index.html.erb
@@ -39,12 +39,7 @@
           <div class="radio" style="padding: 6px 12px; display: inline-block">
             <!-- label --> <%= @bundle.title %>
           </div>
-          <div class="radio" style="padding: 6px 12px; display: inline-block">
-            <span class="sr-only">Download <%= @bundle.title %></span>
-            <%= button_to download_mpl_records_path(@bundle), :method => :get, :class => "btn btn-xs btn-default" do %>
-              Download <span class="sr-only"><%= @bundle.title %></span>
-            <% end %>
-          </div>
+          <%= render partial: "mpl_download", locals: { bundle: @bundle } %>
           <% else %>
             <!-- loop through bundles & use radio buttons -->
             <%= bootstrap_form_tag do |f| %>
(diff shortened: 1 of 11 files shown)


okeefm:
I tested this code in the production Cypress/CVU VM, and it produces downloadable .zip files without throwing an exception for both the 2015 and 2016 bundles

app/views/records/index.html.erb
```diff
@@ -21,38 +21,48 @@
...
-        <% else %>
-          <!-- loop through bundles & use radio buttons -->
-          <%= bootstrap_form_tag do |f| %>
-          <%= f.form_group :bundle_id do %>
+        <div class="col-md-8">
+          <% if Bundle.count == 1 %>
+          <div class="radio" style="padding: 6px 12px;">
+            <!-- label --> <%= @bundle.title %>
+          </div>
+          <% else %>
```
dczulada: Download buttons probably want to be vertically centered on the label

screen shot 2016-11-07 at 2 19 17 pm

Here is an extreme example

screen shot 2016-11-07 at 2 19 05 pm

dczulada:
On one line

screen shot 2016-11-30 at 4 41 46 pm

When screen is made smaller, button moves below text
screen shot 2016-11-30 at 4 41 54 pm

app/views/records/index.html.erb
```diff
@@ -21,38 +21,50 @@
...
+                <div style="display: inline-block">
+                  <%= f.radio_button :bundle_id, bundle.id, label: bundle.title, label_class: "btn btn-checkbox",
+                                     checked: bundle == @bundle %>
+                </div>
+                <div class="radio" style="padding: 6px 12px; display: inline-block">
+                  <span class="sr-only">Download <%= bundle.title %></span>
+                  <%= button_to download_mpl_records_path(bundle), :method => :get, :class => "btn btn-xs btn-default" do %>
+                    Download <span class="sr-only"><%= bundle.title %></span>
+                  <% end %>
+                </div>
```
tstrass: Also there's something off with this form. It works fine if there is only one bundle, but if there are multiple bundles the first Download button acts as a submit for the outer form (defined by the form tag on line 50). So when you click on that button it tries to POST to `bundles/<bundle_id>/records`, which isn't a route that exists. For some reason the second Download button is rendered within its own form (as a child of the line 50 form), so it works as expected.

tstrass: One more thing- we probably don't want the CSS in the html

PR #755, opened 2016-11-18
all product_tests can have duplicate records

dczulada:


okeefm:
Is it that all "`ProductTest`s _will_ have duplicate records", or "all `ProductTest`s _can_ have duplicate records"?

dczulada:
its the second.  bad commit message

okeefm:
Though on second thought this should be coordinated with @lmd59 to make sure it doesn't break her test creation random seed PR (#757)

okeefm:
Going to have to rebase this (and resolve conflicts) because I pulled in #757

PR #756, opened 2016-11-21
Add product filtering for measure selection

rbclark:
Adds the ability for users to filter when they choose to select custom measures. Also adds associated tests for the new functionality.


rbclark:
Thought this was done but I found a few more issues, standby.

rbclark:
Okay this should now be ready for review.

PR #757, opened 2016-12-01
Product random seed

lmd59:
Issue:
Trying to create repeatability by using pseudorandom number generation (with a saved random seed) so that we can reproduce issues

Summary of randomization changes...
 
Product Test (Population Clone Job)
-randomization ids
-date shift
-all controllable demographics randomization (demographics randomizer)
 
Checklist Test
-top measure shuffle
-criteria shuffle (data criteria selector)
 
Filtering Test
-age range (criteria picker)
 
Measure Test
-duplicating records

### DIFF
--- a/app/models/checklist_source_data_criteria.rb
+++ b/app/models/checklist_source_data_criteria.rb
@@ -68,7 +68,7 @@ class ChecklistSourceDataCriteria
 
   def printable_name
     measure = Measure.find_by(_id: measure_id)
-    sdc = measure.hqmf_document[:data_criteria].select { |key, value| key == source_data_criteria }.values.first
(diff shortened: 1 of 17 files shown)


okeefm:
Looks like there are some Rubocop errors in this codebase that need to be fixed. if you run `bundle exec overcommit --sign` (only needs to be run once) then `bundle exec overcommit --run` you'll see the errors being thrown by Travis.

lmd59:
I believe all the rubocop/travis issues have been fixed and pushed. The rubocop autocorrect fixed some issues in files that I hadn't actually touched, but those changes didn't break the unit tests, so I ended up just pushing them up.

PR #758, opened 2016-12-01
full patient download

lmd59:
change measure test to all product tests for full patient download

### DIFF
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -84,7 +84,8 @@ class ProductsController < ApplicationController
 
   # always responds with a zip file of (.qrda.zip files of (qrda category I documents))
   def patients
-    file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda')
+    criteria_list = render_to_string(file: 'checklist_tests/print_criteria.html.erb', layout:'report') if @product.product_tests.checklist_tests.length>0
+    file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda',criteria_list)
     send_data file.read, type: 'application/zip', disposition: 'attachment', filename: "#{@product.name}_#{@product.id}.zip".tr(' ', '_')
   end
 
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -84,7 +84,7 @@ class ProductsController < ApplicationController
 
   # always responds with a zip file of (.qrda.zip files of (qrda category I documents))
   def patients
-    criteria_list = render_to_string(file: 'checklist_tests/print_criteria.html.erb', layout:'report') if @product.product_tests.checklist_tests.length>0
+    criteria_list = (@product.product_tests.checklist_tests.length>0) ? render_to_string(file: 'checklist_tests/print_criteria.html.erb', layout:'report') : nil
     file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda',criteria_list)
     send_data file.read, type: 'application/zip', disposition: 'attachment', filename: "#{@product.name}_#{@product.id}.zip".tr(' ', '_')
   end
--- a/app/controllers/products_controller.rb
+++ b/app/controllers/products_controller.rb
@@ -84,8 +84,9 @@ class ProductsController < ApplicationController
 
   # always responds with a zip file of (.qrda.zip files of (qrda category I documents))
   def patients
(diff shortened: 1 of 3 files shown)


lib/cypress/create_download_zip.rb
```diff
@@ -30,12 +30,24 @@ def self.all_patients
...
+            add_file_to_zip(z, "#{pt.cms_id}_#{pt.id}.#{format}.zip".tr(' ', '_'), pt.patient_archive.read)
+          elsif pt.class == FilteringTest
+            add_file_to_zip(z, "filteringtest_#{pt.cms_id}_#{pt.id}.#{format}.zip".tr(' ', '_'), pt.patient_archive.read) if !added_filter_test
+            added_filter_test = true
+          elsif pt.class == ChecklistTest
+            @product = pt.product
+            zip = Cypress::CreateDownloadZip.create_c1_criteria_zip(@product.product_tests.checklist_tests.first, criteria_list).read
+            add_file_to_zip(z, "checklisttest_#{@product.name}_#{@product.id}_c1_manual_criteria.zip".tr(' ', '_'),zip)
+          else
+            raise "Unknown product test type in download: " >> pt._type
```
okeefm: If you're concatenating strings in Ruby, you'll want to use `+` rather than `<<`

lib/cypress/create_download_zip.rb
```diff
@@ -30,12 +30,24 @@ def self.all_patients
...
-      measure_tests = MeasureTest.where(product_id: product.id)
+    def self.create_total_test_zip(product, format = 'qrda', criteria_list)
+      product_tests = ProductTest.where(product_id: product.id)
       file = Tempfile.new("all-patients-#{Time.now.to_i}")
       Zip::ZipOutputStream.open(file.path) do |z|
-        measure_tests.each do |m|
-          add_file_to_zip(z, "#{m.cms_id}_#{m.id}.#{format}.zip".tr(' ', '_'), m.patient_archive.read)
+        added_filter_test = false
+        product_tests.each do |pt|
+          if pt.class == MeasureTest
```
okeefm: Rather than comparing `class`, the Ruby way to do this is `pt.is_a?`. Alternatively, this control flow seems like it may be better suited to a `case` statement.

app/controllers/products_controller.rb
```diff
@@ -84,7 +84,8 @@ def report
 
   # always responds with a zip file of (.qrda.zip files of (qrda category I documents))
   def patients
-    file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda')
+    criteria_list = (@product.product_tests.checklist_tests.length>0) ? render_to_string(file: 'checklist_tests/print_criteria.html.erb', layout:'report') : nil
+    file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda',criteria_list)
```
okeefm: You'll need a space between the comma after `'qrda'`, and `criteria_list`.

app/controllers/products_controller.rb
```diff
@@ -84,7 +84,8 @@ def report
 
   # always responds with a zip file of (.qrda.zip files of (qrda category I documents))
   def patients
-    file = Cypress::CreateDownloadZip.create_total_test_zip(@product, 'qrda')
+    criteria_list = (@product.product_tests.checklist_tests.length>0) ? render_to_string(file: 'checklist_tests/print_criteria.html.erb', layout:'report') : nil
```
okeefm: You'll also need some spaces here between the `>` symbol and the elements next to it, and the `layout:` and `'report'` elements

lib/cypress/create_download_zip.rb
lmd59: The case statement makes it difficult to eliminate rubocop errors about assignment branch condition size, so I just restructured the way that it was accessing the product tests...

lib/cypress/create_download_zip.rb
```diff
@@ -106,5 +105,24 @@ def self.formatter_for_patients(records, format)
...
+    def self.add_checklist_zips(z, checklist_tests, criteria_list)
+      checklist_tests.each do |pt|
+        p = pt.product
+        file = Cypress::CreateDownloadZip.create_c1_criteria_zip(p.product_tests.checklist_tests.first, criteria_list).read
+        add_file_to_zip(z, "checklisttest_#{p.name}_#{p.id}_c1_manual_criteria.zip".tr(' ', '_'), file)
+      end
+    end
+
+    def self.add_filtering_zips(z, filtering_tests, format)
+      pt = filtering_tests.first
```
okeefm: Do we only add a zip for the first filtering test because they're all the same?

lmd59: Yes. Correct.

PR #760, opened 2016-12-09
Optimize product and vendor load times

rbclark:
This PR has two different parts to it. The first part is some optimizations to make the vendor and product pages load faster. The result of the optimizations in the first commit are as follows:

> Without any optimizations: 112516ms (10340 sql queries in vendors)
> With "includes" optimizations 80057ms (6384 sql queries in vendors)

Additionally the tmp:clear rake task is now enhanced to also rebuild the cache right away after it is cleared. This means that in reality after running tmp:clear (which is run after every upgrade) the load time should be about `7868.5ms` with no sql queries run. I believe the additional time is due to the cache being actually loaded into memory. Subsequent reloads are in the `< 1s` range.

The second part of this is the infinite scrolling. Due to the design of the vendors page it was not possible to implement infinite scrolling on it, the problem being that sorting would no longer work correctly if additional elements were being added as the user scrolled. This problem was mitigated by the above optimizations. For the products page it was possible to implement actual infinite scrolling which helped the performance of that page on a cold cache considerably.

At this point I have not implemented any tests for the infinite scrolling code, if those are desired I can attempt to add those in as well.

@ehathaway The reason I included you on this specifically is due to the product infinite scrolling since it does affect the UI behavior a bit.

### DIFF
--- a/app/controllers/vendors_controller.rb
+++ b/app/controllers/vendors_controller.rb
@@ -9,6 +9,8 @@ class VendorsController < ApplicationController
   add_breadcrumb 'Dashboard', :vendors_path
   add_breadcrumb 'Add Vendor', :new_vendor_path, only: [:new, :create]
 
+  respond_to :js, only: [:show]
+
   def index
     # get all of the vendors that the user can see
     @vendors = Vendor.accessible_by(current_user).order(:updated_at => :desc) # Vendor.accessible_by(current_user).all.order(:updated_at => :desc)
@@ -17,7 +19,7 @@ class VendorsController < ApplicationController
 
   def show
     add_breadcrumb 'Vendor: ' + @vendor.name, :vendor_path
-    @products = Product.where(vendor_id: @vendor.id).order_by(state: 'desc')
+    @products = Product.where(vendor_id: @vendor.id).order_by(state: 'desc').page(params[:page]).per(5)
     respond_with(@vendor)
   end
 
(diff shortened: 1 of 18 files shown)


app/assets/javascripts/infinite-scroll.js.coffee
```diff
@@ -0,0 +1,41 @@
+$ ->
```
okeefm: This is the only `CoffeeScript` file we have in the Cypress code base. For consistency's sake, we may want to rewrite this into plain old JavaScript (sorry...)

app/controllers/vendors_controller.rb
```diff
@@ -9,6 +9,8 @@ class VendorsController < ApplicationController
   add_breadcrumb 'Dashboard', :vendors_path
   add_breadcrumb 'Add Vendor', :new_vendor_path, only: [:new, :create]
 
+  respond_to :js, only: [:show]
```
okeefm: If we set this, `show` will still `respond_to` html, right?

app/assets/javascripts/infinite-scroll.js.coffee
tstrass: I can't personally vouch for any of these but it looks like there are some good tools to convert:
[The try CoffeeScript tab of the CoffeScript website](http://coffeescript.org/)
[js2coffee](http://js2.coffee/)
[decaffeinate](https://github.com/decaffeinate/decaffeinate)

app/controllers/vendors_controller.rb
tstrass: yes

okeefm: 👍

rbclark:
Converted to js, everything seems to still work fine.

PR #763, opened 2016-12-15
Commented out bundle_id in lib/ext/record.rb

okeefm:
To solve a potential `NoMethodError`, `bundle_id` was added to HDS's `Record` model. 

I didn't delete `bundle_id` completely, because we should review whether `bundle_id` should be in HDS's `Record` model in the long term.

This pull request should not get merged in until HDS has been updated.


dczulada:
should be good with a  rebase

PR #765, opened 2016-12-19
New hds

dczulada:


lib/cypress/patient_zipper.rb
```diff
@@ -36,7 +37,12 @@ def initialize(measures, start_time, end_time)
     def export(patient)
       cms_compatible = true if patient.product_test && patient.product_test.product.c3_test
       # qrda version is hard coded right now!!!
```
okeefm: Is this comment still accurate, now that we can switch QRDA versions?

PR #766, opened 2016-12-19
Redo how app config works

tstrass:
This replaces the `APP_CONFIG` hash with the class `Cypress::AppConfig`. The data is still loaded from `cypress.yml`, but instead of a global hash that contains all of the settings, `AppConfig` keeps the data in a cache. It writes to `cypress.yml` and clears the cache in the setter, and reads from the cache and fills it when necessary in the getter. The purpose of this change is to fix the bug where `APP_CONFIG` was not consistently returning the same value in the presence of multiple workers, once a value had been modified.

There are 47 files touched here, but a lot of that is find and replace to change `APP_CONFIG` to `Cypress::AppConfig`. Files with significant changes to be reviewed include:
`lib/cypress/app_config.rb`
`app/controllers/admin/settings_controller.rb`

Note that a few of these changes also switch app config access from using symbols and/or dot notation to using bracket notation and strings. Also note that a couple things are non functional changes to satisfy rubocop rules.

Edit: One additional significant change is adding the usage of the [fakefs](https://github.com/fakefs/fakefs) gem in the tests


lib/cypress/app_config.rb
```diff
@@ -0,0 +1,45 @@
...
+      elsif val.is_a? Symbol
+        yml_text.sub!(sub_string, "#{key}: :#{val}\n")
+      else
+        yml_text.sub!(sub_string, "#{key}: #{val}\n")
+      end
+      File.open("#{Rails.root}/config/cypress.yml", 'w') { |file| file.puts yml_text }
+
+      # Allow setting app config during app init, before Rails.cache exists
+      if Rails.cache.nil?
+        @init_app_config = YAML.load(ERB.new(File.read("#{Rails.root}/config/cypress.yml")).result) if @init_app_config.nil?
```
rbclark: What is the reason for checking if @init_app_config is nil here? Since we just updated the cypress config wouldn't we want to update the @init_app_config value no matter what? Would it hurt preinit to just always hit the filesystem for the latest copy in order to avoid per-thread variables in this specific case?

tstrass: Yeah just put that check to avoid hitting the file system unless we have to, but I'm not opposed to making that change, especially because I don't actually expect anything in `AppConfig` to be set before the cache exists

test/controllers/admin_controller_test.rb
```diff
@@ -7,12 +7,12 @@ class AdminControllerTest < ActionController::TestCase
 
   test 'should get show' do
     Rails.application.config.action_mailer.smtp_settings = {
-      address: APP_CONFIG['mailer_address'],
-      port: APP_CONFIG['mailer_port'],
-      domain: APP_CONFIG['mailer_domain'],
-      user_name: APP_CONFIG['mailer_user_name'],
-      password: APP_CONFIG['mailer_password'],
-      authentication: APP_CONFIG['mailer_authentication']
+      address: Cypress::AppConfig['mailer_address'],
```
okeefm: Do you need the `FakeFS` wrapper here, like you do elsewhere in the testing?

rbclark: Correct, setting values is the only time there is a problem. The actual issue stems from when the tests write to cypress.yml.

okeefm: Ok, sounds good.

PR #770, opened 2016-12-28
Fixing Tests on Travis

rbclark:
Unfortunately you can't test how things build on travis without actually putting in a pull request.

Trying to look into different things that can help the reliability of travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,10 @@ before_install:
   - nvm install stable
   - nvm use stable
   - npm install
+  - mkdir $PWD/travis-phantomjs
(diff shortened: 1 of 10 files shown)


rbclark:
Yep I've unfortunately not gotten back to testing this, probably won't get to until next week.

rbclark:
The final piece of this (and ironically the one that the tests are currently failing on) is going to be done in a separate PR since it is a larger undertaking due to the fact that it is actually a bug in the code, not a bug in the test suite. Once the test suite passes I am comfortable with this being reviewed as is in order to avoid the PR from getting too large.

PR #771, opened 2016-12-28
No manual only

dczulada:


app/views/products/_product_form.html.erb
```diff
@@ -12,15 +12,16 @@
           <fieldset id="certification_options">
             <legend class="control-label">Certification Types</legend>
             <div id="certifications_errors_container"></div>
-            <%= f.form_group help: "You must select C1, C2, or C4 in order to utilize Cypress for testing" do %>
+
+            <%= f.form_group help: "You must select C1, C2, C3, or C4 in order to utilize Cypress for testing" do %>
```
tstrass: I feel like we can remove this help message now that they're all viable by themselves

app/views/products/_product_form.html.erb
```diff
@@ -12,15 +12,16 @@
...
               <% Cypress::AppConfig['certifications'].each do |c, certification| %>
-                <% if c == 'C1' || c == 'C2' || c == 'C4'%>
+                <% if c == 'C1' || c == 'C2' || c == 'C3' || c == 'C4'%>
```
okeefm: Is there any case where we reach this point and `c` is not `C1, C2, C3` or `C4`?

PR #772, opened 2016-12-28
import and export QRDA Cat III R2

dczulada:


dczulada:
This will need HDS updates

okeefm:
Looks like Travis is failing for non-Rubocop reasons

dczulada:
yup...its failing because it needs HDS to be updated first

PR #773, opened 2016-12-29
Remove duplicate declaration of bundle_id

rbclark:
Bundle_id is already declared in the latest version of HDS which we now depend on. Removing the duplicate declaration to get rid of the warning on startup. This happened due to https://github.com/projectcypress/cypress/pull/765 and I unfortunately didn't realize it until a little bit after merging.

### DIFF
--- a/lib/ext/record.rb
+++ b/lib/ext/record.rb
@@ -2,7 +2,6 @@
 
 class Record
   include Mongoid::Document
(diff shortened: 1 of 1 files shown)


rbclark:
Whoops looks like this is a duplicate of https://github.com/projectcypress/cypress/pull/763, favoring that one.

PR #775, opened 2017-01-04
Starred favorites

lmd59:
Adds a favoriting capability to vendor and product list so users can more easily access more important assets. Screen shots are attached below. Review is requested for the user interface which currently uses font awesome stars. 
Additionally, I'm not certain of potential interactions with the current infinite scroll for lists long enough to be paginated.

screen shot 2017-01-03 at 10 40 53 am
screen shot 2017-01-03 at 10 38 43 am

### DIFF
--- a/app/controllers/vendors_controller.rb
+++ b/app/controllers/vendors_controller.rb
@@ -2,14 +2,14 @@ require 'api'
 
 class VendorsController < ApplicationController
   include API::Controller
-  before_action :set_vendor, only: [:show, :update, :destroy, :edit]
-  before_action :authorize_vendor, only: [:show, :update, :destroy, :edit]
+  before_action :set_vendor, only: [:show, :update, :destroy, :edit, :favorite]
+  before_action :authorize_vendor, only: [:show, :update, :destroy, :edit, :favorite]
 
   # breadcrumbs
   add_breadcrumb 'Dashboard', :vendors_path
   add_breadcrumb 'Add Vendor', :new_vendor_path, only: [:new, :create]
 
-  respond_to :js, only: [:show]
+  respond_to :js, only: [:show, :favorite]
 
   def index
     # get all of the vendors that the user can see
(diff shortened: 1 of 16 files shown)


okeefm:
Looks like there are some accessibility issues on the Vendor index page. Travis gives some more details about this; so does running `bundle exec rake cucumber`

okeefm:
I tested this with a Vendor that has 100 Products, and it worked as it should.

app/controllers/vendors_controller.rb
```diff
@@ -71,6 +73,18 @@ def destroy
...
 
+  def favorite
+    user_id = current_user.id
+    deleted_value = @vendor.favorite_user_ids.delete(user_id)
+    @vendor.favorite_user_ids.push(user_id) if deleted_value.nil?
+    @vendor.save!
+    @vendors = Vendor.accessible_by(current_user).order(:updated_at => :desc)
+    @vendors = @vendors.sort_by { |a| (a.favorite_user_ids.include? current_user.id) ? 0 : 1 }
+    respond_with(@vendors.to_a) do |f|
+      f.js {}
```
okeefm: Is this `f.js {}` in the `do` block necessary here? It's in this `favorite` method, but not the equivalent in `products_controller.rb`

lmd59:
Notes from Eny:
favorite sorting (top of list goes missing?)

-bigger star
-remove box outline
-separate favorites area
-bright green (loading color) for star
-star button up top to filter to favorites

Other notes:
-scroll screen to where favorite's new location is (highlight and fade)
-info
-use cases-> how many items likely to be on the list?

lmd59:
Well- cucumber is still having issues for me locally, but since it's passing on Travis, I guess there's not reason not to review.

dczulada:
I keep getting this error when the page needs the infinite scroll.

screen shot 2017-02-09 at 12 17 20 pm

okeefm:
screen shot 2017-02-10 at 11 10 23 am

We should pluralize these based on the number of products in each category. See [this method](http://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize) for some info on how to do that in Rails.

okeefm:
screen shot 2017-02-15 at 8 59 43 am

Is this intended effect when there are favorited vendors, but no non-favorited vendors

PR #778, opened 2017-01-06
Fix bug that caused manual changes to cypress.yml to have no effect

tstrass:
This fixes the bug by reloading the settings in cypress.yml from disc on app init

### DIFF
--- a/lib/cypress/app_config.rb
+++ b/lib/cypress/app_config.rb
@@ -41,5 +41,11 @@ module Cypress
         Rails.cache.delete('config_values')
       end
     end
+
+    def self.refresh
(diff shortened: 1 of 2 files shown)


lib/cypress/app_config.rb
```diff
@@ -41,5 +41,12 @@ def self.[]=(key, val)
         Rails.cache.delete('config_values')
       end
     end
+
+    def self.refresh
+      Rails.cache.delete('config_values')
```
okeefm: Rather than deleting the cache key here, you can just set `force: true` below, so line 47 would look like `Rails.cache.fetch('config_values', force: true) do`

PR #779, opened 2017-01-06
Code lookup feature and Manual Entry Test name change

laclark22:
Created a new feature where you can now click on the valuesets and attritubutes in the record sample form and get a modal pop up with the list of codes associated to that oid. There is a filter functionality for that pop up so you can find the code you need. You can copy and paste any of the codes into the text box. Also changed all of the buttons headers etc. from Manual Entry Test to Record Sample. 

screen shot 2017-01-06 at 3 11 51 pm
screen shot 2017-01-06 at 3 12 01 pm

### DIFF
--- a/app/helpers/checklist_tests_helper.rb
+++ b/app/helpers/checklist_tests_helper.rb
@@ -71,7 +71,7 @@ module ChecklistTestsHelper
 
   def lookup_codevalues(oid, bundle)
     vs = HealthDataStandards::SVS::ValueSet.where(oid: oid)
-    return oid unless vs && vs.first
+    return [] unless vs && vs.first
     vs.first.concepts.map {|con| "#{con.display_name}: #{con.code}"}
   end
 end
--- a/app/helpers/checklist_tests_helper.rb
+++ b/app/helpers/checklist_tests_helper.rb
@@ -59,6 +59,19 @@ module ChecklistTestsHelper
   def lookup_valueset_name(oid)
     vs = HealthDataStandards::SVS::ValueSet.where(oid: oid)
     return oid unless vs && vs.first
-    "#{vs.first.display_name}: #{oid}"
+    "#{vs.first.display_name}"
+  end
+
+
+  def lookup_valueset_oid(oid)
+    vs = HealthDataStandards::SVS::ValueSet.where(oid: oid)
+    return oid unless vs && vs.first
+    "#{oid}"
+  end
+
+  def lookup_codevalues(oid, bundle)
+    vs = HealthDataStandards::SVS::ValueSet.where(oid: oid)
+    return oid unless vs && vs.first
+    vs.first.concepts.map {|con| "#{con.display_name}: #{con.code}"}
   end
 end
--- a/app/views/checklist_tests/_checklist_measure.html.erb
+++ b/app/views/checklist_tests/_checklist_measure.html.erb
@@ -56,13 +56,16 @@
                       <tr>
                       <td> <strong>Valuesets</strong> </td>
                       <td/>
-                      <td><span>
-                        <ul class='list-unstyled'>
+                      <td>
+                      <div type = "button" class ="value-set-group set-menu" data-toggle="modal" data-target="#lookupModal">
+                        <ul class="value-set-list">
                         <% valuessets.each do |vs| %>
-                        <li class="valueset-listitem"><%= "#{lookup_valueset_name(vs)}" %></li>
+                        <li class="value-set-item-header"><%= "#{lookup_valueset_name(vs)}" %></li>
+                        <li class="value-set-item-oid"><%="#{lookup_valueset_oid(vs)}" %></li>
                         <% end %>
                         </ul>
-                      </span></td>
+                        </div>
+                       </td>
                       <td>
                         <% if criteria_field.object.complete?.nil? %>
                           <i class = 'fa fa-fw fa-play-circle text-info invisible' aria-hidden="true"></i>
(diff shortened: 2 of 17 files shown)


app/helpers/checklist_tests_helper.rb
```diff
@@ -59,6 +59,19 @@ def coded_attribute?(criteria)
...
     return oid unless vs && vs.first
-    "#{vs.first.display_name}: #{oid}"
+    "#{vs.first.display_name}"
+  end
+
+
+  def lookup_valueset_oid(oid)
+    vs = HealthDataStandards::SVS::ValueSet.where(oid: oid)
+    return oid unless vs && vs.first
+    "#{oid}"
```
tstrass: I think you probably want this function to do something different, but I'm not sure. Right now it will just return `oid` every time

app/views/checklist_tests/_checklist_measure.html.erb
```diff
@@ -72,13 +75,38 @@
                           <i class = 'fa fa-fw fa-times text-danger' aria-hidden="true"></i>
                         <% end %>
                         <%= criteria_field.text_field :code, hide_label: false %>
+                          <div id="lookupModal" class="modal fade" role="dialog">
```
tstrass: This view is getting pretty bloated. I think it would be best to separate out the modal into a partial.

app/views/checklist_tests/_field_values.html.erb
```diff
@@ -3,21 +3,102 @@
 # local variables
 #
 #   field_values   (Hash)
+#   product_test (ChecklistTest)
```
tstrass: Should add other locals. As far as I can tell there are two others: `negation` and `result`

tstrass: Also since you added the local variable `product_test`, you need to pass in a value for that whenever this partial is rendered. Right now the site breaks if you click the "Download Criteria List" button on the checklist test page, because `product_test` is undefined. You should set the appropriate value for `product_test`[there](https://github.com/projectcypress/cypress/blob/code_lookup/app/views/checklist_tests/print_criteria.html.erb#L48), and then download the criteria list and make sure that looks how it should.

app/views/checklist_tests/_checklist_measure.html.erb
```diff
@@ -72,13 +75,38 @@
...
+                            <div class = "modal-dialog" id ="lookupModal-dialog">
+                              <div class="modal-content" id="lookupModal-content">
+                                <div class= "modal-header" id= "lookupModal-header">
+                                  <button type="button" class="close" id="lookupModal-close" data-dismiss="modal">&times;</button>
+                                    <h4 class="modal-title" id="lookupModal-title"> List of codes</h4>
+                                 </div>
+                                    <div class="modal-body" id="lookupModal-body">
+                                      <input type="text" id="lookupFilter" onkeyup="lookupFunction()" placeholder="Filter codes">
```
tstrass: This input is included as a part of the manual entry form. So if you click into the text box and press enter, it submits the form, which is not the behavior we want, especially because people might not realize that the filter automatically applies as you type. I think you would either need to put this into it's own form or do something like jQuery's [`preventDefault()`](event.preventDefault())

app/views/checklist_tests/_checklist_measure.html.erb
```diff
@@ -72,13 +75,38 @@
...
+                                      <ul class='list-unstyled' id="lookup_codes">
+                                      <% valuessets.each do |vs| %>
+                                        <% lookup_codevalues(vs, product_test.bundle).each do |pair_string| %>
+                                          <li class="valueset-listitem" href="#"><a><%= pair_string %></a></li>
```
tstrass: Each pair is being rendered as a link, but clicking it doesn't do anything. If these are supposed to lead somewhere then you should put in the proper `href`, but if not, then they shouldn't be rendered as links.

app/views/checklist_tests/_field_values.html.erb
```diff
@@ -3,21 +3,102 @@
...
 
 %>
 <% if negation %>
```
tstrass: Here it looks like you have similar html for the modal 3 times, which also looks like the same html for the modal in `_checklist_measure.html.erb` so that can probably be refactored. The best way to do it would be to put the modal in a partial as mentioned in another comment.

dczulada:
still a few places with "Manual".  Change these to "Sample"

screen shot 2017-01-09 at 9 33 52 am

screen shot 2017-01-09 at 9 34 00 am

dczulada:
Also in the "Download Report"
screen shot 2017-01-09 at 9 37 13 am

tstrass:
Closing because the more urgent manual test naming changes were separated out into #781. The review comments here should be addressed in the near future.

app/helpers/checklist_tests_helper.rb
laclark22: That is what I want it to do. Due to the new CSS changes that Eny wanted me to add, the oid and the valueset need to be separated out into different li tags. The other function returned both the oid and the valueset name together so i created a helper function that just displayed the oid.

tstrass: But this function takes `oid` and always just returns `oid`. So it doesn't do anything. If this is in fact the behavior you want, you can just use `oid` directly wherever this function is called.

laclark22: You are correct. Getting rid of that function and just using oid directly.

PR #781, opened 2017-01-09
Rename Manual Entry Test to Sample Record

tstrass:
This consists of a subset of the changes in #779 and all of the changes in #780, plus some fixed tests


dczulada:
I'm sure you already know this now, but the products_helper_test need some assertions to be updated.

PR #782, opened 2017-01-09
one more "manual"

dczulada:


okeefm:
@dczulada there's a "manual" left on line 102 of `app/helpers/test_executions_helper.rb`. Should that still be there?

PR #783, opened 2017-01-09
actually add the format_validators

dczulada:


okeefm:
Looks like this adds some errors, so you may have to fix `test_should_be_able_to_tell_when_potentialy_too_much_data_is_in_documents`

PR #790, opened 2017-01-17
Remove manual

tstrass:
This is a refactor that changes `Manual` to `Checklist` throughout the code. This includes a large number of variable names and method calls. It also includes changing two class names (`C1ManualTask`, `C3ManualTask` changed to `C1ChecklistTask`, `C3ChecklistTask`), which in turn requires a database migration to ensure backwards compatibility. Since Mongoid doesn't have migrations natively, I added the [mongoid_rails_migrations](https://github.com/adacosta/mongoid_rails_migrations) gem. Note I pointed to the `master` branch of that gem, because the author added Mongoid 5 support, but hasn't yet tagged a release that includes it. I inquired [here](https://github.com/adacosta/mongoid_rails_migrations/pull/35).


okeefm:
Running these tests, I get the following error: `fatal:         fatal: incompatible library version - /Users/mokeefe/.rvm/gems/ruby-2.2.5/gems/bcrypt-3.1.11/lib/bcrypt_ext.bundle`

Have you seen this before?

tstrass:
I don't think I have. Does that happen in the unit tests or cucumber tests?

okeefm:
Unit tests; I'm seeing them in (among others) `Admin::UsersControllerTest`

okeefm:
An uninstall/reinstall of Ruby/gems fixed this for me, so I assume it was just a weird thing on my system.

PR #792, opened 2017-02-09
Add a pending state in order to fix a race condition

rbclark:
Often when the test suite ran it would fail because the page was not refreshing properly due to a race condition between the state of the test and some other calls in the code. It was happening because the should_reload_product_test_link? function was being used in conjunction with the task_status to determine the test state. These would get out of sync and cause an incorrect code path to be taken, which would cause the page to stop refreshing even though the tests were not completed. This cleans up that code to make less queries which will cause the state to be more reliable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,9 +40,8 @@ before_script:
   - cp ./.travis/mongoid.yml ./config/mongoid.yml
 script:
   - bundle exec brakeman -qAzw1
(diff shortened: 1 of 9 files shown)


rbclark:
I've run the latest commit about 5 times on travis and have so far been unable to get the cucumber tests to fail. I think I'm happy with this PR at this point.

PR #794, opened 2017-02-15
Make hamburger menu happen earlier in order to stop navigation button s from getting pushed below the header

rbclark:
Thought this was fixed previously however it looks like it still happens when the validation utility button is enabled. Bumping collapse point again.


rbclark:
Ah interesting, apparently the screen size when running in travis CI is smaller and thus it can't find the navigation bar. I am going to tweak the window size unless there are any objections

PR #797, opened 2017-02-20
add indexs to make each modal id unique

dczulada:


app/views/checklist_tests/_field_values.html.erb
```diff
@@ -7,13 +7,13 @@
...
 <% if negation %>
-<div type="button" class="value-set-group set-menu" data-toggle="modal" data-target="#lookupModal-negation">          
+<div type="button" class="value-set-group set-menu" data-toggle="modal" data-target="<%= "#lookupModal-negation#{index}" %>">          
   <ul class="value-set-list">
    <li class="value-set-item-header"><%= "#{lookup_valueset_name(negation)}" %></li>
    <li class="value-set-item-oid"><%= negation %> </li>
  </ul>
 </div>
-<div id="lookupModal-negation" class="modal fade" role="dialog">
+<div id="<%= "lookupModal-negation#{index}" %>" class="modal fade" role="dialog">
```
tstrass: Here we have nearly the same HTML 3 times (lines 16-36, 50-70, 82-102). This should be separated out into a partial template. This is also similar to [_checklist_modal.html.erb](https://github.com/projectcypress/cypress/blob/master/app/views/checklist_tests/_checklist_modal.html.erb), so we may want to try to reuse that before making a new partial.

PR #798, opened 2017-02-20
User activity log

lmd59:
User activity log is found on the user tab. It shows the recent activity done specifically by the logged in user. In particular, it show recent test executions (default ordered by most recent) and related information about them that might be relevant to the user.
Screenshot below:
screen shot 2017-02-20 at 3 43 05 pm

### DIFF
--- a/app/controllers/test_executions_controller.rb
+++ b/app/controllers/test_executions_controller.rb
@@ -18,7 +18,7 @@ class TestExecutionsController < ApplicationController
 
   def create
     authorize! :execute_task, @task.product_test.product.vendor
-    @test_execution = @task.execute(results_params)
+    @test_execution = @task.execute(results_params, current_user)
     respond_with(@test_execution) do |f|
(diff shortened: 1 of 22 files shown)


app/controllers/test_executions_controller.rb
```diff
@@ -18,7 +18,8 @@ def index
 
   def create
     authorize! :execute_task, @task.product_test.product.vendor
-    @test_execution = @task.execute(results_params)
+    user = current_user
+    @test_execution = @task.execute(results_params, current_user)
```
okeefm: should this line use `user` rather than `current_user`, given that we just assigned it above?

okeefm:
Is this the intended effect when a test execution is still executing and you go to the "recent activity" page?
screen shot 2017-02-27 at 2 46 28 pm

And after it finishes:
screen shot 2017-02-27 at 2 46 39 pm

PR #801, opened 2017-02-22
Add mongo 3.4 upgrade to upgrade script

rbclark:
This bumps the version of mongo we are using to mongo 3.4. Overall I have yet to run into any issues experienced by performing this upgrade however we should probably wait to merge this until mongo 3.4 incompatibility is fixed with IBM Power Systems, see the note on [this page](https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/). At this point the fix is already in xenial-proposed but has not yet been rolled out as an official update.

Once ubuntu xenial has version 2.24 of libc6 available in apt we can go ahead and deploy this.

### DIFF
--- a/contrib/upgrade_cypress_30.sh
+++ b/contrib/upgrade_cypress_30.sh
@@ -196,6 +196,15 @@ if [ "$CVU_FOUND" = "true" ]; then
   systemctl restart cypress-validation-utility
 fi
 
+echo "Upgrading Mongo Version..."
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 0C49F3730359A14518585931BC711F9BA15703C6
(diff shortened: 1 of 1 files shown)


rbclark:
This should be ready to go now that glibc is at version 

```
ldd (Ubuntu GLIBC 2.23-0ubuntu7) 2.23
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

okeefm:
The PR says we should be looking for glibc 2.24. Is that where Ubuntu is at?

rbclark:
I'm not sure if I typo'd the version initially or if they backported the fix but the linked page says at least glibc 2.23-0ubuntu5, not sure whats going on there but it looks to me like we're on a version later than the one they list so we should be okay.

PR #802, opened 2017-02-23
Add a field for error backtraces if a test execution errors out on us

okeefm:
This will help with debugging if people report errored test executions, or if we get internal errors during runs of the Measure Evaluator.

### DIFF
--- a/app/models/test_execution.rb
+++ b/app/models/test_execution.rb
@@ -9,6 +9,7 @@ class TestExecution
   field :expected_results, type: Hash
   field :reported_results, type: Hash
   field :qrda_type, type: String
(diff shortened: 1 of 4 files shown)


app/models/test_execution.rb
```diff
@@ -91,8 +92,9 @@ def fail
     save
   end
 
-  def errored
+  def errored(e = nil)
     self.state = :errored
+    self.backtrace = e.message + "\n" + e.backtrace.join("\n")
```
dczulada: should this backtrace also be printed to the log

okeefm: That's a good idea. I'll push up an update.

PR #805, opened 2017-02-27
Added a task to reubild the MPL zips when the tmp cache is rebuilt

okeefm:
When we delete the cache (such as when someone runs the update script), the MPL download zips get deleted too. This task should rebuild them when we rebuild the rest of the cache, so we don't break peoples' servers by trying to build them on the fly.


rbclark:
Looks good to me.

PR #808, opened 2017-03-06
use ajax binding to apply javascript to ajaxComplete and submit

lmd59:
Application-wide binding on buttons submitting (and ajax completing) so that the button's focus is blurred. This allows the user to purposefully put highlighting focus on a button, but if the button is clicked, the focus won't remain there indefinitely. One example is for favorite buttons as seen below, but this is also fixed for other buttons that stay on the same page, such as download buttons.

screen shot 2017-03-06 at 9 26 43 am
screen shot 2017-03-06 at 9 27 25 am


dczulada:
A rebase should now fix the travis errors

PR #813, opened 2017-04-19
Allow duplicate patients to have demographics augmented slightly.

sylinmino:


test/models/record_test.rb
```diff
@@ -27,4 +27,33 @@ def test_record_should_be_able_to_find_calculation_results
     r = Record.find('51703a883054cf84390000d4')
     assert_equal 2, r.calculation_results.count, 'record should have 2 calculated results'
   end
+
+  def assert_record_demographics_equal(r1, r2, assert_true = true)
```
tstrass: The `assert_true` argument and the assert message seem a little awkward to me. I think a slightly cleaner way might be to make it `record_demographics_equal(r1, r2)` and return a bool. So then calling it will be like:
`assert(record_demographics_equal(r1, r2), message)`
or
`assert(!record_demographics_equal(r1, r2), message)`
You could use that to make the messages more specific to which comparison you're making, e.g. `'record demographics should be equal after duplication'` for the one on line 56

lib/ext/record.rb
```diff
@@ -55,4 +55,63 @@ def lookup_provider(include_address = nil)
     return { 'npis' => [provider.npi], 'tins' => [provider.tin], 'addresses' => addresses } if include_address
     { 'npis' => [provider.npi], 'tins' => [provider.tin] }
   end
+
+  def duplicate_randomization(random: Random.new)
+    rec = clone
+    rec = randomize_record_name_or_birth(rec, random: random)
```
tstrass: Two lines above can be chained:
`rec = clone.randomize_record_name_or_birth(rec, random: random)`

sylinmino: Should chain all 3?

tstrass: You could but I was thinking it might be best to leave it as two lines for readability. Just a matter of preference so up to you

sylinmino: I'll chain all 3 if Rubocop yells at me for class length

tstrass: Nevermind first two not chainable! Second two lines could be combined if necessary for rubocop

lib/cypress/app_config.rb
```diff
@@ -40,5 +40,12 @@ def self.[]=(key, val)
       # Allow setting app config during app init, before Rails.cache exists
       clear
     end
+
+    def self.load_yml
+      app_config = YAML.load(ERB.new(File.read("#{Rails.root}/config/cypress.yml")).result)
+      names_config = YAML.load(ERB.new(File.read("#{Rails.root}/config/names.yml")).result)
+      app_config['randomization']['names'] = names_config['names']
```
rbclark: Could this be broken out to into its own constant value in application.rb, perhaps with the name NAME_CONSTANTS? Just like CAT1_CONFIG is [here](https://github.com/projectcypress/cypress/blob/master/config/application.rb#L9). It should then just be accessible by saying `NAME_CONSTANTS['names']`. If you can think of a better variable name than NAME_CONSTANTS thats fine too.

config/names.yml
```diff
@@ -0,0 +1,608 @@
+names:
```
rbclark: I would probably get rid of the top level `names:` declaration so when access I didn't have to say `NAME_CONSTANTS['name']` everytime. Not a necessary change just a thought.

lib/cypress/app_config.rb
sylinmino: So the reason we have it as is is because the names were originally part of the "randomization" of AppConfig. In breaking it up I wanted to bring it back to where it originally was once imported. However, if you think it would be neater to do it with a new constant (maybe NAMES_RANDOM), then let me know and I will get on that.

config/names.yml
sylinmino: If we change it over, then I agree.

lib/cypress/app_config.rb
tstrass: I'm not speaking for Rob but I don't think it's for neatness as much as doing it with a global constant avoids some unnecessary loading of the file (since the values never change at run time). I originally suggested you do it this way because like you said it's where the names were loaded previously, but I have no issue with Rob's method.

tstrass: Also it helps separate out some of the static information in cypress.yml from the configuration values that are mutable

rbclark: Sorry I should've updated this after speaking with Louis directly, when we spoke I explained my reasoning for breaking it out is due to the fact that the app_config interface is being completely deleted in my pull request in favor of a new Settings model. Also I am bit weary of a setup like this in general since if a value was added to cypress.yml later down the road that happened to conflict with the name `app_config['randomization']['names']` then it would be very confusing to the developer trying to figure out why the rest of the values they added to cypress.yml were showing up but the one with the name `app_config['randomization']['names']` wasn't.

PR #814, opened 2017-04-19
Add cypress_3_2 branch to travis

tstrass:
We need the cypress_3_2 travis commit from master on the cypress_3_2 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
 branches:
   only:
   - master
-  - cypress3
+  - cypress_3_2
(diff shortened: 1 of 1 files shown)


okeefm:
This PR is suffering from the `nokogiri` issues fixed in https://github.com/projectcypress/health-data-standards/pull/492

tstrass:
We can't do it this way because our policy only allows rebase which we don't want because that will but cypress_3_2 commits on master. I'll cherry-pick the commit onto cypress_3_2

PR #817, opened 2017-04-21
Bump ruby version to latest patch

rbclark:
It turns out that due to a mistake in https://github.com/projectcypress/cypress-recipe/blob/master/resources/install_app.rb#L9 that all users are already running 2.2.6 anyway, so we may as well develop against it as well.


rbclark:
Is the proper way for me to fix this to target this towards the 3.2 branch or some different way?

okeefm:
If you're targeting 3.2, you should make the Pull Request against the `cypress_3_2` branch

rbclark:
That was more meant in reference to the failing bundler audit since we are on an insecure version of nokogiri on master. Does that mean we are not planning to have another point release for the 3.1 branch? If we are then this PR should be against master since otherwise the upgrade script will try to downgrade everyone who is running the latest version of the cypress vm down to ruby 2.2.5. If we are not going to have another 3.1.x point release though then there is no problem with this being against 3.2.

okeefm:
This pull request can be made against the `cypress_3_2` branch, as we don't anticipate any further `3.1.x` point releases.

PR #818, opened 2017-04-26
Migrate settings from cypress.yml over to database

rbclark:
This shouldn't be merged until https://github.com/projectcypress/cypress/pull/816 is and this has been rebased against it, however in its current state this should be a functional PR

This PR moves all of the user configurable settings from cypress.yml into a settings model. The settings model provides an interface for accessing the settings by calling `Settings.current`. Using the `Settings.current` accessor also provides caching in order to limit the number of DB calls.

The code for handling external email services now alerts the user that they need to restart the server after changing settings. This is due to the fact that directly setting ActionMailer settings from the controller after app initialization causes only the current thread to have the mailer settings updated when running in the production VM.

As for migrations, this PR also adds a rake task `cypress:import:config` which takes the path to a copy of the old config file and a string containing the old environment variable values. It then updates the Settings model to reflect the old cypress configuration.


okeefm:
We should get at least one more positive review before merging this.

PR #824, opened 2017-05-04
Added support for generating MPL download bundle as a background job on MPL page load

rbclark:
Sometimes users manage to get to the MPL download page before the download zip has actually been generated. This adds automatic refreshing while the bundle is generated as well as kicking off the generation if it was not generated on initial bundle load.

### DIFF
--- a/app/controllers/records_controller.rb
+++ b/app/controllers/records_controller.rb
@@ -1,6 +1,8 @@
 class RecordsController < ApplicationController
   before_action :set_record_source, only: [:index, :show, :by_measure]
 
+  respond_to :js, only: [:index]
+
   def index
     unless Bundle.default
       @records = []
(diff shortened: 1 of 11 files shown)


app/controllers/records_controller.rb
```diff
@@ -44,13 +46,14 @@ def download_mpl
     if BSON::ObjectId.legal?(params[:format])
       bundle = Bundle.find(BSON::ObjectId.from_string(params[:format]))
 
-      unless File.exist?(bundle.mpl_path)
-        MplDownloadCreateJob.perform_now(bundle.id.to_s)
+      if bundle.mpl_status != :ready
+        flash[:info] = 'This bundle is currently preparing for download.'
+        redirect_to :back
```
okeefm: Do we want to start the prepare job here if it isn't `:ready` or `:building` (e.g. `:unbuilt`)?

rbclark: If we just called mpl_prepare instead of mpl_status here then that would be taken care of. The main reason I did not is because mpl_prepare is already called from the mpl_download page.

okeefm: Ok, I see that now. Looks good to me then.

PR #800, opened 2017-02-22
Move to Mongo 3.4 on travis CI

rbclark:
This PR bumps travis CI to use mongo 3.4 instead of 3.0 for testing.

PR #825, opened 2017-05-09
Update Travis

ssayer:
Move to new travis "trusty tahr" environement and update dependencies

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
+dist: trusty
+sudo: false
 language: ruby
 cache:
   - bundler
   - node_modules
-sudo: false
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,6 @@ notifications:
     secure: NfcNb3LCyd7/ACYc6mDoBJZlBFIV5wKMMWs0On2DNeaDoD/EKCbQRQQVOHEvfsaqOM+5zmi9giYaV7cV+Gl+zfehZ/Q69x89YjWWVFeAp5QGDzoCnM7oaaTXygla52sfcL9Oi3rLgTWyjqMSS7tM2DW5d/V+iEjyilmavsUvRbU=
 before_install:
   - . $HOME/.nvm/nvm.sh
-  - nvm install stable
-  - nvm use stable
-  - npm install
   - mkdir $PWD/travis-phantomjs
   - curl -sSL https://github.com/Medium/phantomjs/releases/download/v2.1.1/phantomjs-2.1.1-linux-x86_64.tar.bz2 -o $PWD/travis-phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2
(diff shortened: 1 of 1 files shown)


rbclark:
It looks like using the phantomJS version that comes with with trusty is still a bit old, it will probably still work better if we use 2.1.1 directly.

ssayer:
Are we still getting that weird Mongo JS loading failure?

rbclark:
Yeah unfortunately those errors are still popping up. Do you know what version of mongo the trusty beta is running? I believe we are planning to target 3.4 for the next release, is that the version they are running or do we need to add the 3.4 apt sources?

ssayer:
I don't. The old travis image used 3.2, which necessitated us downgrading to 3.0. It look like this passed so I would assume it's using 3.4?

rbclark:
It looks like it is running on 3.2 still, I actually had a PR in before to bump to 3.4 #800 and it got merged and somehow reverted.

okeefm:
#800 was merged into `cypress_3_2`, not master, so we're still building on master with 3.0

rbclark:
Oh in that case this is gonna cause a bit of merge conflict for the person merging 3.2, perhaps this should be put against cypress_3.2?

ssayer:
Changed base to 3.2. I mostly did this as an experiment, but if you want this merged this into 3.2, go ahead and deconflict.

rbclark:
`Scenario: Can Multi Upload to the Same Task on a Measure Test Multiple Times` is having intermittent failures again as a result of this PR, it could be the new container infrastructure is a lot slower to run than the old travis which is causing it to timeout. If we merge this in I worry we will start to get a lot of intermittent test failures again. There will probably need to be some investigation into why exactly this has popped back up.

rbclark:
Small followup on the previous error, it seems that the random failures may be happening due to the trusty container infrastructure running significantly slower, the old jobs seem to take between 17-20 minutes on precise but take 33-36 minutes on trusty. I'm going to look into bumping the timeout to see if that helps. Also going to look into whether the page still has a pending refresh based on the html output of this failing test.

rbclark:
Sorry for hijacking this PR, unfortunately the issue is made much worse by this version of travis so testing on this branch is the most convenient place to do so. I have tried a few changes and I hope they help improve the reliability.

ssayer:
No problem, I just wanted to do a few experiments (some of which failed, like switching from poltergeist to webkit). If it's being useful, I'm glad it's living on!

rbclark:
I believe the multi upload is stable at this point. Unfortunately other tests are still intermittently failing, perhaps simply due to how slow it seems our tests our running on the trusty beta. I would suggest we hold off on merging this until we have more information about why they are running so badly, I have sent an email to travis support to see if they have any input.

rbclark:
I have moved all of the fixes for this PR in to a new pull request specific to those fixes instead of them being tied in with the travis update.

PR #827, opened 2017-05-11
Make changes to calling go export with init func

mattrbianchi:


lib/cypress/patient_zipper.rb
```diff
@@ -66,7 +66,7 @@ def self.zip(file, patients, format)
                   else
                     Cypress::HTMLExporter.new(mes, sd, ed)
                   end
-
+
```
okeefm: nitpicky, but can you get rid of this whitespace?

PR #830, opened 2017-05-19
Add logic to expect augmented duplicates if they are detected in a va…

dczulada:
…lidation

### DIFF
--- a/app/models/product_test.rb
+++ b/app/models/product_test.rb
@@ -18,6 +18,8 @@ class ProductTest
 
   has_many :records, :dependent => :destroy, :foreign_key => :test_id
 
+  field :augmented_records, type: Array, default: []
+
   field :expected_results, type: Hash
   # this the hqmf id of the measure
   field :measure_ids, type: Array
@@ -89,10 +91,7 @@ class ProductTest
       prng = Random.new(rand_seed.to_i)
       ids = results.where('value.IPP' => { '$gt' => 0 }).collect { |pc| pc.value.patient_id }
       unless ids.nil? || ids.empty?
-        dups = records.find(ids)
-        (prng.rand(3) + 1).times do
-          recs << dups.sample(random: prng).duplicate_randomization(random: prng)
-        end
+        recs = sample_and_duplicate_records(recs, ids, random: prng)
       end
     end
     Cypress::PatientZipper.zip(file, recs, :qrda)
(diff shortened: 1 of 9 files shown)


lib/validators/calculating_augmentated_records.rb
```diff
@@ -0,0 +1,73 @@
...
+        original_results = QME::PatientCache.where('value.medical_record_id' => mrn, 'value.test_id' => @test_id,
+                                                   'value.measure_id' => measure.hqmf_id, 'value.sub_id' => measure.sub_id).first
+        options[:population_ids] = measure.population_ids
+        passed = compare_results(original_results, results, options, passed)
+      end
+      record.destroy
+      passed
+    end
+
+    # Functions related to expected measure calculation results
```
dczulada: Do these three functions (augmented_sup_val_expected, reported_in_expected_range, augmented_record_in_population) belong here, or in another class that is more related to the expected_results_validator.

ljades: My intention was to move all functionality related to expecting augmented duplicates into one class. I do see though that there is enough to be considered distinct about these two sections of the class to separate them. I will get on that.

app/models/product_test.rb
```diff
@@ -104,6 +103,23 @@ def archive_records
     save
   end
 
+  def sample_and_duplicate_records(recs, ids, random: Random.new)
+    car = ::Validators::CalculatingAugmentedRecords.new(measures, records, id)
```
dczulada: I believe the records in the initialize for Smoking Gun Validator is only being used for the name comparision. This isn't needed for the validate_calcuated_results methods. Could possibly be changed to

car = ::Validators::CalculatingAugmentedRecords.new(measures, [], id)

app/models/product_test.rb
```diff
@@ -104,6 +103,23 @@ def archive_records
...
+    dups = records.find(ids)
+
+    (random.rand(3) + 1).times do
+      prng_repeat = Random.new(rand_seed.to_i)
+      dup_rec, rec_augments, old_rec = dups.sample(random: random).duplicate_randomization(random: prng_repeat)
+      if car.validate_calculated_results(dup_rec, 'effective_date' => effective_date)
+        augmented_records << rec_augments
+        recs << dup_rec
+      else
+        recs << old_rec
```
dczulada: Should something be added to "augmented_records" to indicate that the patient is "duplicate" but no other changes have been made?  Maybe just at the record_id?

dczulada:
PR looks good. Approved.   I 'technically' can't review this since I opened the PR on behalf of Louis.

PR #833, opened 2017-05-25
Fix measure search filtering

rbclark:
Currently the measure search loses focus after every ajax query. This change was made in order to fix the favorite buttons from holding focus. This filters those buttons down to only remove focus if the active element is a button. Also I have added some checks to stop multiple AJAX queries from happening at once.


lmd59:
Absolutely, the blur action was meant to fix button focus issues. Changes look good to me.

PR #837, opened 2017-05-31
If the entry is negated, there is an NA_VALUESET codeset and there is…

dczulada:
… only one code, find a code to use for calcuation

### DIFF
--- a/lib/validators/calculating_smoking_gun_validator.rb
+++ b/lib/validators/calculating_smoking_gun_validator.rb
@@ -42,6 +42,13 @@ module Validators
       record = HealthDataStandards::Import::Cat1::PatientImporter.instance.parse_cat1(doc)
       record.test_id = te.id
       record.medical_record_number = rand(1_000_000_000_000_000)
+      record.entries.each do |entry|
+        # If the entry is negated, there is a NA codeset and there is only one code, find a code to use for calcuation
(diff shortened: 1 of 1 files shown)


lib/validators/calculating_smoking_gun_validator.rb
```diff
@@ -42,6 +42,13 @@ def parse_and_save_record(doc, te, options)
       record = HealthDataStandards::Import::Cat1::PatientImporter.instance.parse_cat1(doc)
       record.test_id = te.id
       record.medical_record_number = rand(1_000_000_000_000_000)
+      record.entries.each do |entry|
+        # If the entry is negated, there is a NA codeset and there is only one code, find a code to use for calcuation
+        if entry.negationInd == true && entry.codes.key?('NA_VALUESET') && entry.codes.size == 1
```
okeefm: Do we actually need to do the `entry.negationInd == true` here (e.g. do we actually care about a direct comparison to `true`), or can we just test for truthiness with `entry.negationInd`?

okeefm: @dczulada

dczulada: yes, we want to make sure that negationInd == true and not negationInd == false

PR #838, opened 2017-06-06
Fix C1 Task Test hangups

ljades:


app/models/c1_task.rb
```diff
@@ -3,7 +3,6 @@
```
tstrass: I think you can get rid of these `require`s now

ljades: Not sure why the comment isn't shown as outdated after I rebased, but done.

tstrass: The line I commented on didn't change

ljades: Gotcha.

PR #849, opened 2017-06-13
Go Integration

tstrass:
Use `go-cda-tools` gem for import and export.


okeefm:
Looks like there are some issues the test suite is picking up with `export_with_ffi`

tstrass:
@okeefm that commit should fix it. had to run a `bundle update` for the go-cda-tools gem.
There will still be a rubocop error for that one commented line, but @dczulada is changing that soon.

tstrass:
New Travis failures are now due to out-of-date linux binary in go-cda-tools. That update should be in soon and then we can update the dependency here again.

lib/cypress/patient_zipper.rb
```diff
@@ -22,8 +22,10 @@ def export(patient)
   end
 
   class QRDAExporter
-    C3EXPORTER = HealthDataStandards::Export::Cat1.new('r3')
-    C3_1EXPORTER = HealthDataStandards::Export::Cat1.new('r3_1')
+    C3EXPORTER = GoCDATools::Export::GoExporter.instance
+    C3_1EXPORTER = GoCDATools::Export::GoExporter.instance
```
tstrass: I think we actually do need them to be 3 different instances of the exporter. not positive but I think we should leave it like this for now. And I think Matt's future performance improvements will make this more efficient.

okeefm:
Do you want to squash these commits down into 1/a few commits?

tstrass:
Not yet. It looks like we want to bump go-cda-tools again so i'll take care of it

tstrass:
🎉

PR #850, opened 2017-06-14
Go import

dczulada:


lib/cypress/demographics_randomizer.rb
```diff
@@ -39,7 +39,6 @@ def self.randomize_address(record)
       address = Address.new
       address.use = 'HP'
       address.street = ["#{Faker::Address.street_address} #{Faker::Address.street_suffix}"]
-      #address.street.push(Faker::Address.secondary_address) if [true, false].sample
```
tstrass: can we remove this?

lib/validators/go_import.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    def update_entries(record, refs)
+      record.entries.each do |entry|
+        if refs.include?(entry.cda_identifier.extension)
+          entry._id = BSON::ObjectId.from_string(entry.cda_identifier.extension)
+        else
+          entry._id = entry.cda_identifier._id
+        end
+        # If the entry is negated, there is a NA codeset and there is only one code, find a code to use for calcuation
+        if entry.negationInd == true && entry.codes.key?('NA_VALUESET') && entry.codes.size == 1
```
tstrass: `entry.negationInd == true` can just be `entry.negationInd`

PR #851, opened 2017-06-14
Clinical Randomization: split by a given date or types

okeefm:
* Split by date: Randomizer picks a date that has at least 1 entry on
both sides of it, then splits on that date
* Split by type: Randomizer splits types between two records, so all of
one type are in one record

Continued work on clinical randomization

Initial tests for clinical randomization

Tests for clinical randomization complete

Fully tested clinical randomizer

Clean up/add comments, and add split_by_type

Integrate clinical randomizer with population duplication

remove the clinically duplicated patient from the test deck before duplication

Deal with record entries that aren't named the same thing as their embedded relations

### DIFF
--- a/app/models/product_test.rb
+++ b/app/models/product_test.rb
@@ -107,6 +107,8 @@ class ProductTest
     car = ::Validators::CalculatingAugmentedRecords.new(measures, [], id)
     dups = records.find(ids)
 
+    recs, dups = randomize_clinical_data(recs, dups, random)
+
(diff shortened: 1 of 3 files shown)


dczulada:
Looks good.  Behaves how I would expect it to.  Would it be possible to export the insurance provider in both files?  In order to be a valid QRDA document (that passes schematron) there needs to be a least one entry.  Some people might cry foul if we don't do this.

screen shot 2017-06-14 at 10 51 23 pm

PR #857, opened 2017-06-17
Add use of load_measures_and_value_sets

mattrbianchi:


lib/cypress/patient_zipper.rb
```diff
@@ -40,20 +38,18 @@ def initialize(measures, start_time, end_time)
         @valueset_json_map[bundle.id] = value_sets.to_json
       end
       @measures_json = @measures.to_json
+      CAT_EXPORTER.load_measures_and_value_sets(@measures_json, @valueset_json_map[@measures.first.bundle_id])
```
dczulada: @valueset_json_map is probably no longer needed.  can pass in HealthDataStandards::SVS::ValueSet.where(:oid.in => @measures.map(&:oids).flatten, bundle_id: @measures.first.bundle_id).to_json directly

lib/cypress/patient_zipper.rb
```diff
@@ -40,20 +38,18 @@ def initialize(measures, start_time, end_time)
...
     end
 
     def export(patient)
       cms_compatibility = patient.product_test && patient.product_test.product.c3_test
       case patient.bundle.qrda_version
```
dczulada: No longer need case pass this in as a parameter patient.bundle.qrda_version

okeefm:
Looks like this should pass once the relevant `go-cda-tools` branch is merged into master

okeefm:
If you update your Gemfile to the latest `go-cda-tools`, we should be good to pull this in.

lib/cypress/patient_zipper.rb
```diff
@@ -22,9 +22,7 @@ def export(patient)
   end
 
   class QRDAExporter
-    C3EXPORTER = GoCDATools::Export::GoExporter.instance
-    C3_1EXPORTER = GoCDATools::Export::GoExporter.instance
-    C4EXPORTER = GoCDATools::Export::GoExporter.instance
+    CAT_EXPORTER = GoCDATools::Export::GoExporter.instance
```
tstrass: Can we call this `QRDA_EXPORTER` instead? CAT on it's own doesn't really make sense

PR #858, opened 2017-06-19
Replace c1

dczulada:


dczulada:
Original C1 Testing

before

Edit Test to replace Encounter, Performed with Medication, Allergy

edit

Updated Test

after

app/helpers/checklist_tests_helper.rb
```diff
@@ -27,6 +27,35 @@ def fulfills_reference(measure, referenced_criteria)
     end
   end
 
+  def availble_data_criteria(measure, criteria)
```
okeefm: `availble` should be `available`

app/views/checklist_tests/_checklist_measure.html.erb
```diff
@@ -107,6 +105,9 @@
                           <% else %>
                             <td/>
                           <% end %>
+                          <td class='show-me' style='display: none;'>
+                            <%= criteria_field.select :replacement_data_criteria, availble_data_criteria(measure, criteria), :selected => criteria_field.object.source_data_criteria, class: 'show-me' %>
```
okeefm: `availble` should be changed to `available` here as well

app/helpers/checklist_tests_helper.rb
dczulada: Another example of where going to the Derek Zoolander center for kids who can't read good really isn't paying off.

okeefm:
The only last comment I have here is that it's not obvious to me what the "Toggle Negation" switch does. Does that toggle the negation of the entire valueset or just a single code? If so, we may want to grey out the code selector when the toggle is on, or change the text that says "negate entire valueset", so it's more obvious what the switch is doing.

dczulada:
If I click the toggle now, it will negate the entire valueset
valueset

If I click the toggle now, it will negate an individual code
single

PR #860, opened 2017-06-20
Fix bugs in demographics randomization code

ljades:


okeefm:
Can you give some detail as to what bugs this should fix?

dczulada:
Yes.  The bug in question.

Original (pre-demographics randomization)

expected - gender (M:0, F:3)
reported - gender (M:1, F:2) - Error M should be 0, F should be 3 

Original (post-demographics randomization, pre bug fix)
expected - gender (M:0, F:3)
Patient 1 was copied twice (and changed from F to M)
expected - gender (M:1, F:2) or (M:0, F:3)
reported - gender (M:2, F:1) - No error returned

Since the original file was copied twice, the validator was allowing the demographics to be altered twice (even though the copied patient should only be reported once)

app/models/product_test.rb
```diff
@@ -108,10 +108,9 @@ def sample_and_duplicate_records(recs, ids, random: Random.new)
     dups = records.find(ids)
 
     recs, dups = randomize_clinical_data(recs, dups, random)
-
-    (random.rand(3) + 1).times do
+    dups.sample(random.rand(3) + 1, random: random).each do |rec|
```
tstrass: `random.rand(3) + 1` is equivalent to `random.rand(1..4)`

lib/validators/expected_supplemental_results.rb
```diff
@@ -4,14 +4,19 @@ module Validators
   class ExpectedSupplementalResults < QrdaFileValidator
     include Validators::Validator
 
+    def initialize(file_name)
+      @file_name = file_name
+      super()
```
tstrass: should remove `()` when no arguments to method call, just a ruby convention

app/models/product_test.rb
ljades: Wouldn't it be `random.rand(1..3)`?

okeefm: I think @ljades is right, as `random.rand(3)` will only give you integers `0`, `1`, and `2`

tstrass: If what you want is a random number in the range [1, 4]
`random.rand(3) + 1` and `random.rand(1..4)` will both do that.

If that's not what you want, then what you have here is wrong.

okeefm: What we want here is `1`, `2`, or `3`. The way to do that is `random.rand(1..3)`.

As a side note, `random.rand(3) + 1` will also give you `1`, `2`, or `3`. `random.rand(1..4)` will include `4` as well.

tstrass: My mistake you are correct @okeefm. 
So yeah @ljades like he said we want `random.rand(1..3)`

test/helpers/test_executions_helper_test.rb
```diff
@@ -112,6 +112,60 @@ def test_get_select_history_message
     assert_includes msg, '(passing)'
   end
 
+  def test_population_errors
+    errors = []
```
tstrass: @ljades would it make sense to just make one errors array at put it in the setup for this test class?

PR #861, opened 2017-06-23
Some negations need to use specific codes for consistent results

dczulada:


dczulada:
@okeefm What are your thoughts on this approach?

okeefm:
This strategy looks good to me.

okeefm:
@dczulada let me know when this PR is updated with the updated codes, and I'll review/merge.

dczulada:
@okeefm good to go

PR #864, opened 2017-06-29
Move away from ruby dockerfile and instead use one provided by phusion

rbclark:
This provides us with a docker container that has a proper init setup meaning that we can actually start multiple processes in the one container as needed.

This actually has to be merged before I can upgrade the staging server since I will need to be able to pull the image off DockerHub.


okeefm:
I'm running a bundle import task in the Docker machine; I'll merge once that finishes successfully.

PR #865, opened 2017-06-29
Allow reporting in the expected results validator using HL7 or CMS payer codes

ljades:


lib/validators/expected_supplemental_results.rb
```diff
@@ -25,27 +25,43 @@ def check_supplemental_data_matches_pop_sums(report_sup_val, keys_and_ids, expec
...
                     !CalculatingAugmentedResults.augmented_sup_val_expected?(options['task'], keys_and_ids, code,
-                                                                             { expect: expect_sup_val[code], report: report_sup_val[code] },
+                                                                             { expect: expect_sup_val[code], report: report_val },
                                                                              modified_population_labels)
         add_sup_data_error(keys_and_ids, code, 0, report_val)
       end
     end
 
+    def get_untranslated_code(code, keys_and_ids)
+      return code if keys_and_ids[:sup_key] != 'PAYER'
```
okeefm: Can we pass in solely the `keys_and_ids[:sup_key]` instead of the whole hash?

lib/validators/expected_supplemental_results.rb
```diff
@@ -25,27 +25,43 @@ def check_supplemental_data_matches_pop_sums(report_sup_val, keys_and_ids, expec
...
         report_val = report_sup_val.nil? ? 0 : report_sup_val[code]
         add_sup_data_error(keys_and_ids, code, expect_val, report_val)
       end
     end
 
+    def get_translated_report_val(report_sup_val, code, keys_and_ids)
+      return nil if report_sup_val.nil?
+      payer = APP_CONSTANTS['randomization']['payers'].find { |p| p['code'].to_s == code }
+      code_cms = payer ? payer['codeCMS'] : ''
+      return report_sup_val[code] if keys_and_ids[:sup_key] != 'PAYER' || !report_sup_val[code].nil? || report_sup_val[code_cms].nil?
```
okeefm: Similarly here, can we pass in solely `keys_and_ids[:sup_key]`?

PR #870, opened 2017-07-12
Use the criteria_field's index to uniquely identify toggle elements

dczulada:


dczulada:
before

before

after

after

okeefm:
After toggling a negation and then saving the test, I got the following error: 
screen shot 2017-07-14 at 11 27 40 am

dczulada:
@okeefm Since Transfers are only attributes.  They cannot be swapped.

from this
was2

to this
couldbe

PR #871, opened 2017-07-13
Limited history

lmd59:
Limited user history to 15 most recent entries on the user page. (Includes pagination with progress bar on load.)
http://localhost:3000/users/edit

screen shot 2017-07-13 at 11 19 23 am


app/controllers/registrations_controller.rb
```diff
@@ -19,7 +21,14 @@ def edit
   protected
 
   def load_test_executions
-    @test_executions = current_user.test_executions
+    # if session[:stored_tests]
```
okeefm: It might make sense to get rid of some of the commented-out code here

PR #873, opened 2017-07-13
Add a CCN to the Provider when generating a new Provider

okeefm:
Do Not Merge until HDS PR is merged.

### DIFF
--- a/lib/ext/provider.rb
+++ b/lib/ext/provider.rb
@@ -19,6 +19,7 @@ class Provider
     prov.family_name = NAMES_RANDOM['last'].sample
     prov.npi = NpiGenerator.generate
     prov.tin = rand.to_s[2..10]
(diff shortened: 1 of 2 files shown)


lib/ext/provider.rb
```diff
@@ -19,6 +19,7 @@ def self.generate_provider(options = {})
     prov.family_name = NAMES_RANDOM['last'].sample
     prov.npi = NpiGenerator.generate
     prov.tin = rand.to_s[2..10]
+    prov.ccn = rand(1..66).to_s.ljust(2, '0') + rand(1..9899).to_s
```
tstrass: could the left side of the addition be simplified to `rand(10..66).to_s` or is that different from what you want?

okeefm: It is actually different, but because I screwed up. I should have used `rjust` here. Thanks for the catch.

tstrass: Ah ok that makes sense

PR #876, opened 2017-07-20
Added upgrade script for Cypress 3.2

rbclark:
Note that the upgrade script will not work until the Cypress 3.2.0 final release is tagged. For testing just change the Cypress version tag at the top of the file to the version you wish to have it pull down (master is also acceptable as a tag)


rbclark:
One last thing that I just realized would be good to add to this is the ability for us to lock the version of mongo to 3.4.5. Currently we force downgrade to that version but we don't actually lock it to that version meaning if users decide to run an `apt upgrade` then it will break their system. It looks like chef has support for this so it may not hurt to attempt to replicate the chef functionality here as well.

okeefm:
Looks like something like `sudo apt-mark hold mongodb-org mongodb-org-mongos mongodb-org-server mongodb-org-shell mongodb-org-tools` would do the trick, though I haven't tested it yet.

rbclark:
Yep thats the conclusion I came to as well, also have not tested yet. Will do later today.

PR #885, opened 2017-07-24
use jquery to show and hide labels

dczulada:


dczulada:
The show and hide was not working correctly

app/assets/javascripts/checklist_measures.js
```diff
@@ -7,11 +7,11 @@ function lookupLabelFunction(index) {
     input2 = document.getElementById("vs"+index);
```
okeefm: This variable doesn't need to exist anymore. This is what's breaking Travis.

PR #887, opened 2017-07-25
Iterate over changed criteria in reverse order, previously newly crea…

dczulada:
…ted criteria where being evaluated


dczulada:
If you try to update a lot of criteria

before

The change_criteria starts to evalute the newly added criteria

after

Changing the order to start from the end of the list fixes this.

after2

okeefm:
How does this fix actually fix this? If I were to start changing criteria in reverse order, would it break like before?

PR #888, opened 2017-07-25
Should be able to press Edit Test after upload without refresh.

dczulada:


dczulada:
Same approach as here https://github.com/projectcypress/cypress/commit/8dca2d827003aa84d5c0099ace1d387998c7599d

PR #891, opened 2017-07-26
Update version number to 3.2.0

dczulada:


dczulada:
This should be the last PR pulled in before tagging.

PR #893, opened 2017-07-26
Only use MPL ids if they are in the format of MPL Records

dczulada:


dczulada:
Currently, the array includes two types of values

...

["098718e0-4d42-0135-8680-20999b0ed66f",
 "78214fda23d953c002f02e81d98e674c"]

only the ones formatted like "098718e0-4d42-0135-8680-20999b0ed66f" are in the records table

screen shot 2017-07-26 at 1 45 26 pm

The others are not


screen shot 2017-07-26 at 1 46 30 pm

PR #895, opened 2017-07-27
Fix time displayed in Most Recent executions dropdown on tests

rbclark:
Currently everywhere else uses the local_time gem except for this dropdown, adding to the dropdown as well to standardize.

### DIFF
--- a/app/helpers/test_executions_results_helper.rb
+++ b/app/helpers/test_executions_results_helper.rb
@@ -4,7 +4,7 @@ module TestExecutionsResultsHelper
   def get_select_history_message(execution, is_most_recent)
     msg = ''
     msg << 'Most Recent - ' if is_most_recent
-    msg << execution.created_at.in_time_zone.strftime('%B %e, %Y %l:%M%P')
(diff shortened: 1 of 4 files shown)


dczulada:
https://github.com/projectcypress/cypress/commit/bbacc1e0e373a62df3b6c663cc72ea10fdddbd50

app/helpers/test_executions_results_helper.rb
```diff
@@ -4,7 +4,7 @@ module TestExecutionsResultsHelper
   def get_select_history_message(execution, is_most_recent)
     msg = ''
     msg << 'Most Recent - ' if is_most_recent
-    msg << execution.created_at.in_time_zone.strftime('%B %e, %Y %l:%M%P')
+    msg << local_time(execution.created_at)
```
okeefm: Looks like `local_time` isn't a helper method here

rbclark: Yeah, interestingly it works when you run the app itself but I cannot figure out how to get it to be included from the test suite.

PR #897, opened 2017-08-02
A referenced id should be able to follow any reasonable convention, n…

dczulada:
…ot just a BSON objectId


dczulada:
This will also need a go-cda-tools bundle update

lib/cypress/go_import.rb
```diff
@@ -1,35 +1,11 @@
 module Cypress
   module GoImport
-    def self.update_conditions(record)
```
okeefm: Do we not need to update conditions anymore? Or are we doing this on the `cdatools` side?

dczulada: https://github.com/projectcypress/go-cda-tools/pull/28

okeefm: 👍

PR #901, opened 2017-08-07
Add hdd usage alerts

rbclark:
Users are now alerted when they have between 80% and 100% of their disk full.

Also this fixes miscellaneous rubocop errors.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery :with => :exception, :unless => -> { request.format.json? || request.format.xml? }
 
   before_action :restrict_basic_auth, :authenticate_user!, :check_bundle_installed, :check_backend_jobs,
-                :check_remaining_disk_space, :except => %i[page_not_found server_error]
+                :check_remaining_disk_space, except: %i(page_not_found server_error)
   around_action :catch_not_found
 
   rescue_from CanCan::AccessDenied do |exception|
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -44,16 +45,29 @@ def any_bundle
...
   end
 
+  def check_remaining_disk_space
+    Vmstat.snapshot.disks.each do |disk|
+      disk_percentage = 100 - ((disk.available_blocks / disk.total_blocks.to_f) * 100).round
+      disk_full_msg = "Your disk is #{disk_percentage}% full, please check your remaining hard disk space and " \
+                      'clean up old tests in order to ensure continued stable operation of the Cypress application.'
+      if disk_percentage >= 95
+        flash.now[:disk_space_danger] = disk_full_msg
+      elsif disk_percentage >= 80
```
dczulada: Since this will pop up on each click...maybe we should have the warning be at 90%.  Or can we make the warnings only show up once per uesr session?

rbclark: It will take a bit more work to get it to show up once per session. I could perhaps only show it after the sign in route is hit, or I would also be fine with just making the limit 90%

app/controllers/application_controller.rb
```diff
@@ -44,16 +45,27 @@ def any_bundle
     bundle
   end
 
+  def check_remaining_disk_space
+    Vmstat.snapshot.disks.each do |disk|
+      disk_percentage = 100 - ((disk.available_blocks / disk.total_blocks.to_f) * 100).round
+      disk_full_msg = "Your disk is #{disk_percentage}% full, please check your remaining hard disk space and " \
+                      'clean up old tests in order to ensure continued stable operation of the Cypress application.'
+      if disk_percentage >= 95
```
dczulada: flash.now[:disk_space_danger] = disk_full_msg if disk_percentage >= 95

PR #903, opened 2017-08-08
Update version number

dczulada:


dczulada:
Just to have around and ready prior to tagging

PR #911, opened 2017-09-27
Allow for tests to be shifted into the actual reporting period.

dczulada:
Default behavior is to use the date in the bundle.


okeefm:
We should work up some tests for calculation on shifting dates, and probably some Cucumber tests for the UI.

test/unit/lib/validators/measure_period_validator_test.rb
```diff
@@ -11,16 +11,42 @@ def setup
...
+    @test_execution.task.product_test.product.shift_records = true
+    # measure_period_start and effective dates are set using milliseconds (below are for 2015)
+    @test_execution.task.bundle.measure_period_start = 1_420_070_400
+    @test_execution.task.bundle.effective_date = 1_451_606_399
+    @validator.validate(file, 'test_execution' => @test_execution)
+    errors = @validator.errors
+    assert_equal 2, errors.length, 'should have 2 errors for the invalid reporting period'
+  end
+
+  def test_file_with_unshift_mp
```
ljades: What is the difference between this and the previous test? The function name and their related files don't seem to give the best picture of what distinguishes them and makes them expect different results.

dczulada:
This ticket address the feature requested in https://jira.mitre.org/browse/CYPRESS-45.

This feature allows for a user (vendor or ATL) to choose whether or not to shift test records in to actual reporting period.  This feature reuses the shift_dates functionality in HDS to shift the patient records.  The amount of time for the shift is determined by the bundle (and specified in the bundle configurtion in cypress.yml).  The amount of time is specified in years (e.g., the 2016.0.0 bundle has records from the 2015 calendar year, and the acutal reporting period is for 2017.  therefore the dates need to be shifted forward 2 years).

All references for effective_date and measure_period_start should reference the product, and not the bundle.  If shift_dates is not specified, the products measure_period_start will return the bundles measure_period_start.  The effective_date can be determined relative to the measure_period_start.

In addition to shifting records, all measure dependent validations needs to use the shifts as well

- Cat 3 Calculator
- Measure Period Validator


**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-45

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

dczulada:
I need to update the reference in Clinical Randomizer to reference the products effective dates.

app/models/product.rb
```diff
@@ -33,6 +34,16 @@ class Product
   validate :valid_measure_ids?
   validates :vendor, presence: true
 
+  def effective_date
+    # If selected, move effective date forward to the end of the actual reporting period
+    shift_records ? (Time.at(measure_period_start).in_time_zone + 1.year - 1.second).to_i : bundle.effective_date
+  end
+
+  def measure_period_start
```
okeefm: I'd suggest putting this method before `effective_date` for clarity into where the start_date_offset is coming in.

PR #912, opened 2017-09-27
Don't treat the device order template different than other templates

dczulada:


dczulada:
This was a workaround for when a negated device, order did not have a template id.
This has since been fixed with QRDA Cat I R3.1.
Measure using QRDA Cat I R3 were advised to report negated device, applied instead of negated device, orders

This code is no longer needed

PR #919, opened 2017-10-05
Add deprecation to bundles

rbclark:
This PR adds support for deprecation of Bundles in Cypress, a feature which allows users to mark a bundle as no longer supported, while still allowing users who are using the bundle to finish up their work before complete removal. It is still currently a WIP, however since it affects some of the core operation of Cypress I am putting it up for review even though the tests are not yet written. 

Bundles now must be accessed with the Bundles.available scope which only shows bundles that are not deprecated. The admin panel is the only exception to this since the admin panel must show deprecated bundles so users are able to eventually delete them. 

In an attempt to reuse as much code as possible, the ChecklistTests now also have a background worker run when they are created in order to setup the ChecklistTest patient archive. This does not contain the `checklist_test.html` due to the fact that the `render_to_string` method which is used to generate the checklist test is only accessible from a Controller, [at least until Rails 5](https://stackoverflow.com/a/42924760).

Also in an attempt to reuse as much code as possible, remove_modal has been renamed to action_modal in order to support both the remove and deprecate functionality.

**To Do:**
- [x] Verify that the user is be able to install a different version of the bundle (e.g., installing the 2016.0.2 bundle after deprecating the 2016.0.0 bundle) - This requires https://github.com/projectcypress/health-data-standards/pull/546 to be merged before it will work.
- [x] Verify that the Download Criteria list functionality still works properly under normal operation if a user changes which measures are associated with this product. (This seems to be acting strange even on master)

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-24

**Reviewer 1:**

Name: Dave Czulada

**Reviewer 2:**

Name: Louis Ades


dczulada:
We shouldn't need to worry about version numbers being the same.

PR #920, opened 2017-10-06
Calculating Augmented Results needs to track positive and negative di…

dczulada:
…ffs separately

This fixes a bug identified on the Cypress external JIRA https://oncprojectracking.healthit.gov/support/browse/CYPRESS-1217.

The problem here is the "expect_max_diff" variable in CalculatingAugmentedResults can wind up cancelling itself out in the scenario where randomizations oppose one another.

For example:

Joe
	original gender: M
	duplicate gender: F

Jill
	original gender: F
	duplicate gender: M

In this case, the expect_max_diff (for F) will +1 and then -1

**Submitter:**
- [x] External ticket for this PR: CYPRESS-1217
- [x] External ticket links to this PR: CYPRESS-1217

**Reviewer 1:**

Name: Louis

**Reviewer 2:**

Name: Laura

### DIFF
--- a/test/unit/lib/validators/expected_results_validator_test.rb
+++ b/test/unit/lib/validators/expected_results_validator_test.rb
@@ -91,6 +91,8 @@ class ExpectedResultsValidatorTest < ActiveSupport::TestCase
     @record1.save!
     @validator.validate(file, 'task' => @task)
     errors = @validator.errors
+    # Note: the 16 errors here are the same as the test_validate_missing_supplemental_data test (we are using the same task)
+    # The importance here is that no new errors are introduced when changing the gender count in accordance with the augmented records
     assert_equal 16, errors.length, 'should error on missing supplemental data' # 13 errors related to pop sums
     errors.each { |e| (assert_equal :result_validation, e.validator_type) }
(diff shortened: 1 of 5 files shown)


test/unit/lib/validators/expected_results_validator_test.rb
```diff
@@ -60,6 +84,75 @@ def test_validate_missing_supplemental_data
...
+                      population_id: 'E83A6DA6-D34E-4107-8431-2FB2C86738C7',
+                      data_type: 'SEX',
+                      code: 'F',
+                      expected_value: 3,
+                      reported_value: 1 }
+
+    # The should be error messages when reported value is outside of the expected range.  In this example, the range is 2-3.
+    assert_equal 3, errors.count { |e| e.validator_type == :result_validation && e.error_details == error_details }
+  end
+
```
ljades: There should be one last test case for a good result due to the augmented patients.

ljades: Actually, scratch that. These are meant to be "good" cases. If they could have a comment added to them mentioning that, since they read like "bad" cases since they expect errors--they just expect no more errors than before.

PR #924, opened 2017-10-11
If a filtering test for a product already has an archive, use it and …

dczulada:
…copy the augmented records.

**Submitter:**
- [x] Internal ticket for this PR: https://oncprojectracking.healthit.gov/support/browse/CYPRESS-1226
- [x] Internal ticket links to this PR https://oncprojectracking.healthit.gov/support/browse/CYPRESS-1226

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/app/models/filtering_test.rb
+++ b/app/models/filtering_test.rb
@@ -15,8 +15,12 @@ class FilteringTest < ProductTest
   def archive_records
     # If a filtering test for this product already has an archive, use it and copy the augmented records.
     # Otherwise, create them.
-    if !product.product_tests.filtering_tests.where(:patient_archive.ne => nil).empty?
-      first_filter_test = product.product_tests.filtering_tests.find_by(:patient_archive.ne => nil)
+    first_filter_test = begin
+                          product.product_tests.filtering_tests.find_by(:patient_archive.ne => nil)
+                        rescue
+                          nil
+                        end
+    if first_filter_test
       self.augmented_records = first_filter_test.augmented_records
       self.patient_archive = first_filter_test.patient_archive
     else
--- a/app/models/filtering_test.rb
+++ b/app/models/filtering_test.rb
@@ -12,6 +12,18 @@ class FilteringTest < ProductTest
     save
   end
 
+  def archive_records
+    # If a filtering test for this product already has an archive, use it and copy the augmented records.
+    # Otherwise, create them.
+    if !product.product_tests.filtering_tests.where(:patient_archive.ne => nil).empty?
+      first_filter_test = product.product_tests.filtering_tests.find_by(:patient_archive.ne => nil)
+      self.augmented_records = first_filter_test.augmented_records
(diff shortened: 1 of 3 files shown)


app/models/filtering_test.rb
```diff
@@ -12,6 +12,22 @@ def create_tasks
     save
   end
 
+  def archive_records
+    # If a filtering test for this product already has an archive, use it and copy the augmented records.
+    # Otherwise, create them.
+    first_filter_test = begin
+                          product.product_tests.filtering_tests.find_by(:patient_archive.ne => nil)
```
okeefm: This might be a place where the Ruby 2.3 `&.` (safe navigation) operator is cleaner than a `begin/rescue/end` block. See @rbclark's changes in https://github.com/projectcypress/cypress-validation-utility/blob/master/app/controllers/uploads_controller.rb#L48, but basically if it would throw a `NoMethodError` (e.g. the `find_by` returned `nil`), it just returns `nil` instead.

okeefm: `product&.product_tests&.filtering_tests&.find_by(:patient_archive.ne => nil)` for example

rbclark: The rescue in this case is actually due to the `find_by` throwing an exception. It is expected that all of the other parts will exist, there is just no way to get find_by to return nil for a single case, short of setting it to do so for the whole app.

okeefm: Ok, that's fine then. As much as I don't like the construct I see the reasoning behind it.

rbclark:
Even with the fix I have still been able to get different test decks when running multiple background workers. I have been creating a C4 Test with 1 measure and then clicking "Download QRDA Category I (.zip)" for each test. I ran the test twice, once with 1 background worker and another time with 4 background workers

With 1 background worker:
```
rbclark ~/Downloads $ diff CMS30v5_59de7f6bc60ac13f99d2fb50.qrda CMS30v5_59de7f6bc60ac13f99d2fb53.qrda
rbclark ~/Downloads $ diff CMS30v5_59de7f6bc60ac13f99d2fb50.qrda CMS30v5_59de7f6bc60ac13f99d2fb53.qrda
rbclark ~/Downloads $ diff CMS30v5_59de7f6bc60ac13f99d2fb50.qrda CMS30v5_59de7f6bc60ac13f99d2fb56.qrda
rbclark ~/Downloads $ diff CMS30v5_59de7f6bc60ac13f99d2fb50.qrda CMS30v5_59de7f6bc60ac13f99d2fb59.qrda
```

With 4 background workers:
```
rbclark ~/Downloads $ diff CMS30v5_59de8134c60ac140b013de84.qrda CMS30v5_59de8134c60ac140b013de8a.qrda | wc -l
    1256
```

rbclark:
After the last commit I have not been able to reproduce this issue, should now be resolved.

PR #935, opened 2017-10-31
Code quality

ljades:
This PR is now ready to be reviewed.

It is a compilation of static code analysis revisions on the Cypress codebase.

**Submitter:**
- [x] Internal ticket for this PR:

**Reviewer 1:**

Name:
- [x] The tests appropriately test the new code, including edge cases (N/A)

**Reviewer 2:**

Name: Robert Clark

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -15,14 +15,7 @@ module Admin
 
     def update
       @user = User.find(params[:id])
-      @user.roles = []
-      @user.add_role params[:role]
-      @user.email = params[:user][:email] if params[:user] && params[:user][:email]
-      assignments = params[:assignments].values if params[:assignments]
-      (assignments || []).each do |ass|
-        @user.add_role(ass[:role], Vendor.find(ass[:vendor_id]))
-      end
-      @user.save
+      @user.assign_roles_and_email(params)
       flash[:notice] = 'Successfully updated user.'
       redirect_to_admin
     end
@@ -41,10 +34,7 @@ module Admin
 
     def unlock
       @user = User.find(params[:id])
-      @user.locked_at = nil
-      @user.failed_attempts = 0
-      @user.unlock_token = nil
-      @user.save
+      @user.unlock
       redirect_to_admin
     end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -109,12 +109,12 @@ class ApplicationController < ActionController::Base
 
   def set_product_test
     @product_test = params[:product_test_id] ? ProductTest.find(params[:product_test_id]) : ProductTest.find(params[:id])
-    @title = "#{@product_test.product.name} C1 Record Sample" if @product_test.is_a?(ChecklistTest)
+    @title = "#{@product_test.product_name} C1 Record Sample" if @product_test.is_a?(ChecklistTest)
   end
 
   def set_task
     @task = params[:task_id] ? Task.find(params[:task_id]) : Task.find(params[:id])
-    @title = "#{@task.product_test.product.name} #{@task.product_test.cms_id} #{@task._type.titleize}"
+    @title = "#{@task.product_test.product_name} #{@task.product_test_cms_id} #{@task._type.titleize}"
   end
 
   def set_test_execution
--- a/app/controllers/test_executions_controller.rb
+++ b/app/controllers/test_executions_controller.rb
@@ -20,6 +20,7 @@ class TestExecutionsController < ApplicationController
   def create
     authorize! :execute_task, @task.product_test.product.vendor
     @test_execution = @task.execute(results_params, current_user)
+    @curr_task = Task.find(params[:task_id])
     respond_with(@test_execution) do |f|
       if @task.is_a? C1ChecklistTask
         f.html { redirect_to product_checklist_test_path(@task.product_test.product, @task.product_test) }
@@ -80,10 +81,11 @@ class TestExecutionsController < ApplicationController
   end
 
   def add_breadcrumbs
+    product = @product_test.product
     add_breadcrumb 'Dashboard', :vendors_path
-    add_breadcrumb 'Vendor: ' + @product_test.product.vendor.name, vendor_path(@product_test.product.vendor)
-    add_breadcrumb 'Product: ' + @product_test.product.name, vendor_product_path(@product_test.product.vendor, @product_test.product)
-    add_breadcrumb 'Record Sample Test', product_checklist_test_path(@product_test.product, @product_test) if @product_test.is_a? ChecklistTest
+    add_breadcrumb 'Vendor: ' + product.vendor_name, vendor_path(product.vendor)
+    add_breadcrumb 'Product: ' + product.name, vendor_product_path(product.vendor, product)
+    add_breadcrumb 'Record Sample Test', product_checklist_test_path(product, @product_test) if @product_test.is_a? ChecklistTest
     add_test_execution_breadcrumb
   end
 
@@ -102,9 +104,5 @@ class TestExecutionsController < ApplicationController
     elsif params[:test_execution]
       params[:test_execution][:results]
     end
-  end
-
-  def file_params
-    params.require(:file_name)
   end
 end
--- a/app/controllers/vendors_controller.rb
+++ b/app/controllers/vendors_controller.rb
@@ -19,14 +19,13 @@ class VendorsController < ApplicationController
 
   def show
     add_breadcrumb 'Vendor: ' + @vendor.name, :vendor_path
-    product_arr = Product.where(vendor_id: @vendor.id).order_by(state: 'desc').sort_by { |a| (a.favorite_user_ids.include? current_user.id) ? 0 : 1 }
-    @products_fav = product_arr.select { |p| p.favorite_user_ids.include? current_user.id }
+    @products_fav = @vendor.favorite_products(current_user)
 
     # paginate non-favorites
-    products_nonfav = product_arr.select { |p| !(p.favorite_user_ids.include? current_user.id) }
+    products_nonfav = @vendor.products.ordered_for_vendors.select { |p| !(p.favorite_user_ids.include? current_user.id) }
     @nonfav_count = products_nonfav.count
     @products_nonfav = Kaminari.paginate_array(products_nonfav).page(params[:page]).per(5)
-    @products = product_arr
+    @products = @vendor.products.ordered_for_vendors
     respond_with(@vendor)
   end
 
--- a/app/models/checklist_test.rb
+++ b/app/models/checklist_test.rb
@@ -125,9 +125,7 @@ class ChecklistTest < ProductTest
     patient_archive
   end
 
-  def cms_id_from_measure_id(measure_id)
-    Measure.find_by(_id: measure_id).cms_id
-  rescue
-    'cms id not found'
+  def most_recent_task_execution_incomplete?
+    tasks.any? && tasks[0].most_recent_execution && tasks[0].most_recent_execution.incomplete?
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -92,7 +92,7 @@ class User
 
   def associate_points_of_contact
     if Settings.current.auto_associate_pocs
-      Vendor.where('points_of_contact.email' => email).each do |vendor|
+      Vendor.where('points_of_contact.email' => email).by_updated_at.each do |vendor|
         add_role :vendor, vendor
       end
     end
@@ -105,5 +105,23 @@ class User
 
   def user_role?(*args)
     has_role?(*args) || Settings.current.ignore_roles
+  end
+
+  def assign_roles_and_email(params)
+    self.roles = []
+    add_role params[:role]
+    self.email = params[:user][:email] if params[:user] && params[:user][:email]
+    assignments = params[:assignments].values if params[:assignments]
+    (assignments || []).each do |ass|
+      add_role(ass[:role], Vendor.find(ass[:vendor_id]))
+    end
+    save
+  end
+
+  def unlock
+    self.locked_at = nil
+    self.failed_attempts = 0
+    self.unlock_token = nil
+    save
   end
 end
--- a/app/models/vendor.rb
+++ b/app/models/vendor.rb
@@ -4,7 +4,7 @@ class Vendor
   include Mongoid::Timestamps
   resourcify
 
-  default_scope -> { order(:updated_at => :desc) }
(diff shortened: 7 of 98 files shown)


app/views/application/_product_status_table.html.erb
```diff
@@ -42,7 +42,7 @@
       <th scope="col" class="c1-heading <%= 'hide-test' unless product.c2_test %>">QRDA-I</th>
       <th scope="col" class="c2-heading">QRDA-III</th>
       <th scope="col" class="c3-heading <%= 'hide-test' unless product.c1_test %>">Sample</th>
-      <th scope="col" class="c3-heading <%= 'hide-test' unless (product.c1_test and product.c2_test) or product.c3_test %>">QRDA-I</th>
```
dczulada: Going through this, i don't think this code needs to go into the model.  Instead

`<th scope="col" class="c1-heading">Sample</th>
      <th scope="col" class="c1-heading">QRDA-I</th>
      <th scope="col" class="c2-heading">QRDA-III</th>
      <th scope="col" class="c3-heading <%= 'hide-test' unless (product.c1_test and product.c3_test) %>">Sample</th>
      <th scope="col" class="c3-heading <%= 'hide-test' unless product.c3_test %>">QRDA-I</th>
      <th scope="col" class="c3-heading <%= 'hide-test' unless product.c3_test %>">QRDA-III</th>
      <th scope="col" class="c4-heading">QRDA-I</th>
      <th scope="col" class="c4-heading">QRDA-III</th>`

app/controllers/test_executions_controller.rb
```diff
@@ -82,7 +83,7 @@ def set_product_test_from_task
   def add_breadcrumbs
     add_breadcrumb 'Dashboard', :vendors_path
     add_breadcrumb 'Vendor: ' + @product_test.product.vendor.name, vendor_path(@product_test.product.vendor)
-    add_breadcrumb 'Product: ' + @product_test.product.name, vendor_product_path(@product_test.product.vendor, @product_test.product)
+    add_breadcrumb 'Product: ' + @product_test.product_name, vendor_product_path(@product_test.product.vendor, @product_test.product)
```
ssayer: There's still a lot of trainwrecking in the method. I'm not sure the best way to solve it, but I think that's what making things so unreadable.

app/views/application/_product_status_table.html.erb
ljades: So you think the logic for this whole section should be changes? I can definitely do that, but unless I'm not catching something, that seems to be the situation.

app/controllers/vendors_controller.rb
```diff
@@ -19,8 +19,9 @@ def index
 
   def show
     add_breadcrumb 'Vendor: ' + @vendor.name, :vendor_path
-    product_arr = Product.where(vendor_id: @vendor.id).order_by(state: 'desc').sort_by { |a| (a.favorite_user_ids.include? current_user.id) ? 0 : 1 }
-    @products_fav = product_arr.select { |p| p.favorite_user_ids.include? current_user.id }
+    product_arr = Product.where(vendor_id: @vendor.id).by_updated_at.order_by(state: 'desc')
```
ssayer: Can't this be pushed down tot he model level?

app/models/vendor.rb
```diff
@@ -40,7 +40,7 @@ def self.accessible_by(user)
   # calls on product, product tests, etc, all the way down and does them manually. This means that
   # if any of those structures change then this code will need to be updated accordingly.
   def destroy
-    product_tests = ProductTest.where(:product_id.in => product_ids)
+    product_tests = ProductTest.where(:product_id.in => product_ids).by_updated_at
```
ssayer: Can we review the code for commonly used where conditions and make them into scopes? This doesn't have to be done as part of this PR, but it's something to look at.

lib/cypress/measure_evaluator.rb
```diff
@@ -67,13 +67,13 @@ def wait_for_calculations
...
         sleep(2)
-        continuing = ProductTest.where(state: 'pending', product: product).count > 0
+        continuing = ProductTest.where(state: 'pending', product: product).by_updated_at.count > 0
       end
       @logger.info 'done'
     end
 
     def upload_all_cat3s
-      upload_cat3s(Product.where(name: 'MeasureEvaluationProduct').first.product_tests)
+      upload_cat3s(Product.where(name: 'MeasureEvaluationProduct').by_updated_at.first.product_tests)
```
ssayer: by_updated_at.first is kinda wonky. Shouldn't that be a scope called "most_recent' or something like that? I see a bunch of places where we're doing something like that

app/views/application/_product_status_table.html.erb
ssayer: Like I said, I'm not sure. Talk to @dczulada about it, he has the final say

dczulada: After a future look.  It only needs to be this.  The tests will already be "hidden" if the product is not a c3.  The only case that now needs a check is the Record Sample.  This is only included if C1 is true.

```
<th scope="col" class="c1-heading">Sample</th>
      <th scope="col" class="c1-heading">QRDA-I</th>
      <th scope="col" class="c2-heading">QRDA-III</th>
      <th scope="col" class="c3-heading <%= 'hide-test' unless product.c1_test">Sample</th>
      <th scope="col" class="c3-heading">QRDA-I</th>
      <th scope="col" class="c3-heading">QRDA-III</th>
      <th scope="col" class="c4-heading">QRDA-I</th>
      <th scope="col" class="c4-heading">QRDA-III</th>
```

app/controllers/test_executions_controller.rb
ljades: I'm gonna push a first round of changes before I fix this, just because it's still unsure how to go about the fix. 

I will start by changing the above line too though, since it's strange that rails_best_practices didn't call it out. It will at least make it a bit more consistent.

ljades: Actually, I'm going to try substituting local variables to make it more readable.

Gemfile.lock
```diff
@@ -517,6 +529,7 @@ DEPENDENCIES
   pry-nav
   quality-measure-engine!
   rails (~> 4.2.7)
+  rails_best_practices
```
okeefm: It looks like you removed this from the `Gemfile`, but didn't rerun `bundle install`

ljades: Whoops. My b. Updated now.

app/controllers/application_controller.rb
```diff
@@ -109,12 +109,12 @@ def set_product
...
   def set_product_test
     @product_test = params[:product_test_id] ? ProductTest.find(params[:product_test_id]) : ProductTest.find(params[:id])
-    @title = "#{@product_test.product.name} C1 Record Sample" if @product_test.is_a?(ChecklistTest)
+    @title = "#{@product_test.product_name} C1 Record Sample" if @product_test.is_a?(ChecklistTest)
   end
 
   def set_task
     @task = params[:task_id] ? Task.find(params[:task_id]) : Task.find(params[:id])
-    @title = "#{@task.product_test.product.name} #{@task.product_test.cms_id} #{@task._type.titleize}"
+    @title = "#{@task.product_test.product_name} #{@task.product_test_cms_id} #{@task._type.titleize}"
```
okeefm: TIL about Rails Prefixes

app/models/vendor.rb
```diff
@@ -73,4 +73,17 @@ def status
...
+
+  def header_fields?
+    url? || address? || !points_of_contact.empty?
+  end
+
+  def product_arr
+    Product.where(vendor_id: id).by_updated_at.order_by(state: 'desc')
+  end
+
+  def favorite_products(current_user)
```
rbclark: We have favorites for both products and vendors but they are both done very differently. Currently the vendors are all gathered in 1 array and then filtered in the view, and the products are selected from the DB twice. Is this just due to us only updating 1 of the methods and not the other?

app/models/vendor.rb
```diff
@@ -73,4 +73,17 @@ def status
...
+    product_arr_sorted = product_arr.sort_by { |a| (a.favorite_user_ids.include? current_user.id) ? 0 : 1 }
+    product_arr_sorted.select { |p| p.favorite_user_ids.include? current_user.id }
```
rbclark: I am having trouble understanding the purpose of the previous 2 lines. Mongo supports directly querying a database for the contents of an array, so replacing both the above lines with `product_arr.where(favorite_user_ids: current_user.id)` should actually work fine. I tested this locally with 2 separate users favoriting the same product and it seems to work fine to me. As you can see below even though favorite_user_ids has 2 elements in it, it still returned the product with multiple users favoriting properly.

```
product_arr.where(favorite_user_ids: current_user.id).first
=> #<Product _id: 5a21b467c60ac152505cbd7e, created_at: 2017-12-01 19:58:31 UTC, updated_at: 2017-12-01 20:24:13 UTC, vendor_id: BSON::ObjectId('5a21b3cec60ac151e4eda020'), bundle_id: BSON::ObjectId('5a1c3eb1c60ac172eeccafc6'), name: "test", version: "", description: "", c1_test: true, c2_test: false, c3_test: false, c4_test: false, randomize_records: true, duplicate_records: true, shift_records: false, allow_duplicate_names: false, measure_selection: "custom", measure_ids: ["40280381-4BE2-53B3-014B-E6419B2C033E"], favorite_user_ids: [BSON::ObjectId('5a1c3e86c60ac170d8a0f5d0'), BSON::ObjectId('5a21b95ec60ac152505cbda1')]>
```

app/models/vendor.rb
```diff
@@ -73,4 +73,17 @@ def status
       product_counts.key?(product_state) ? product_counts[product_state].count : 0
     end
   end
+
+  def header_fields?
+    url? || address? || !points_of_contact.empty?
+  end
+
+  def product_arr
+    Product.where(vendor_id: id).by_updated_at.order_by(state: 'desc')
```
rbclark: There is already an association between vendors and product, it makes sense here to just say `products.by_updated_at.order_by(state: 'desc')`. Also this whole method seems like it could be replaced with a scope on the product called `ordered_for_vendors`.

app/views/admin/show.html.erb
```diff
@@ -15,22 +15,22 @@
   </ul>
 
   <div id="application_settings">
-    <%= render partial: "show_settings", locals: { banner_message: banner_message, warning_message: warning_message, banner: banner, smtp_settings: smtp_settings, default_url_options: default_url_options, mode: mode, mode_settings: mode_settings, debug_features: debug_features, server_needs_restart: server_needs_restart } %>
+    <%= render 'show_settings', :banner_message => banner_message, :warning_message => warning_message, :banner => banner, :smtp_settings => smtp_settings, :default_url_options => default_url_options, :mode => mode, :mode_settings => mode_settings, :debug_features => debug_features, :server_needs_restart => server_needs_restart %>
```
rbclark: Is there any way to reuse the hash built in the model containing all of the current settings here?

lib/ext/bundle.rb
```diff
@@ -26,7 +26,7 @@ def deprecate
 
   def destroy
     results.destroy
-    Product.where(bundle_id: id).destroy_all
+    Product.where(bundle_id: id).by_updated_at.destroy_all
```
rbclark: The sort is unnecessary here since we are just destroying them anyway.

test/controllers/records_controller_test.rb
```diff
@@ -5,7 +5,7 @@ class RecordsControllerTest < ActionController::TestCase
   setup do
     collection_fixtures('bundles', 'records', 'vendors', 'products', 'product_tests', 'tasks', 'users', 'measures', 'roles')
     @vendor = Vendor.find(EHR1)
-    @first_product = @vendor.products.where(name: 'Vendor 1 Product 1').find('4f57a88a1d41c851eb000004')
+    @first_product = @vendor.products.where(name: 'Vendor 1 Product 1').by_updated_at.find('4f57a88a1d41c851eb000004')
```
rbclark: I'm not sure the by_updated_at is necessary here either since find only returns 1 object, ditto on all the other finds in the controller tests.

app/controllers/admin/users_controller.rb
```diff
@@ -15,14 +15,7 @@ def show
...
       @user = User.find(params[:id])
-      @user.roles = []
-      @user.add_role params[:role]
-      @user.email = params[:user][:email] if params[:user] && params[:user][:email]
-      assignments = params[:assignments].values if params[:assignments]
-      (assignments || []).each do |ass|
-        @user.add_role(ass[:role], Vendor.find(ass[:vendor_id]))
-      end
-      @user.save
+      @user.update(params)
```
ssayer: Can we confirm we're not clobbering some existing method called "update" by doing this?

ljades: Updated it. Should be fine now.

app/models/user.rb
```diff
@@ -92,7 +92,7 @@ def password_complexity
 
   def associate_points_of_contact
     if Settings.current.auto_associate_pocs
-      Vendor.where('points_of_contact.email' => email).each do |vendor|
+      Vendor.where('points_of_contact.email' => email).by_updated_at.each do |vendor|
```
rbclark: This seems like it is another case of not needing the `by_updated_at` since I don't believe the order of adding a role matters.

app/models/checklist_test.rb
```diff
@@ -125,9 +125,7 @@ def archive_records
     patient_archive
   end
 
-  def cms_id_from_measure_id(measure_id)
-    Measure.find_by(_id: measure_id).cms_id
-  rescue
-    'cms id not found'
+  def most_recent_task_execution_incomplete?
+    tasks.any? && tasks[0].most_recent_execution && tasks[0].most_recent_execution.incomplete?
```
rbclark: Is there any reason we are using tasks[0] here instead of tasks.first?

app/views/test_executions/_expected_results.html.erb
```diff
@@ -1,29 +1,29 @@
...
 
-<% elsif @task.product_test.expected_results %>
+<% elsif task.product_test_expected_results %>
 <h1>Expected Aggregate Results</h1>
 <table class="table table-condensed table-hover">
   <thead>
     <tr>
       <th scope="col">Measure Name</th>
-      <% population_keys = @task.product_test.expected_results.first[1].population_ids.keys %>
+      <% population_keys = task.product_test_expected_results.first[1].population_ids.keys %>
```
rbclark: There is actually a `.second` method in ruby, just FYI.

ssayer:
Whoever is going to merge this, please squash the commits

PR #937, opened 2017-11-08
Build release packages with pkgr

rbclark:
This PR changes us over to have automated builds of Cypress using Pkgr. Pkgr is a platform that uses the Heroku buildpacks to create a packaged version of Cypress which is able to be installed using `apt-get`, instead of requiring a manual build process. The cypress-recipe Chef cookbook is still necessary, however almost all of the previous dependencies have been removed and it is now a very barebones cookbook to deploy mongo and the Cypress package. The upgrade script has been significantly cleaned up as well (included in this PR), and now simply requires some basic package upgrades and service restarts.

I kept the upgrade script changes in this PR very barebones for now, as I did not want to bring over a large amount of potentially unneeded steps (for example I was unsure of the usefulness of running `rake tmp:clear` command on upgrade going forward). When we are testing for the next upgrade release we should make sure that these steps are all we need. I have tested them against basic changes and they work fine.

When this PR is merged we will have to tweak the deploy process slightly. We will now need to have a stable branch and we will need to tag and push to this branch whenever there is a new stable release. we can also have a beta or RC branch if we so choose. The reason for this is the hosted Pkgr service, http://packager.io, builds every time you push to a branch. If we were to build against master then every time someone merged a PR a new package would be created and any user who ran `sudo apt-get upgrade on their box would get an unstable version.

This also fixes an issue I found during testing where the auto_approve was not wrapped in a proc and therefore was only evaluating on application load, not every time a new user was being created. This meant that if a user had auto_approve off and then turned it on it would not apply until the application was restarted.

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-2

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


rbclark:
Closing since this has been merged directly into the cypress_v4 branch.

PR #949, opened 2017-12-08
Enforce 2014 certification edition settings in the model

rbclark:
This includes the changes required to get debug_mode working properly when the 2014 certification edition is selected. In the model we now check if the 2014 cert_edition is selected and if it is then we use a before_save hook to force the values to be false for duplicate_records and c4_test.

**Submitter:**
- [x] Internal ticket for this PR: Still related to https://jira.mitre.org/browse/CYPRESS-107

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/app/assets/javascripts/measure_selection.js
+++ b/app/assets/javascripts/measure_selection.js
@@ -127,7 +127,6 @@ function UpdateMeasureSet(bundle_id) {
 // as the second parameter. True means disabled and false means enabled.
 function setCheckboxDisabled(element, state) {
   $(element).prop("disabled", state);
-
   if (state) {
     $(element).parent().addClass('disabled');
     $(element).prop("checked", false);
@@ -269,6 +268,7 @@ ready_run_once = function() {
       }
       else if (edition == '2015') {
         setCheckboxDisabled('#product_duplicate_records', false);
+        $('#product_duplicate_records').prop("checked", true);
         setCheckboxDisabled('#product_c4_test', false);
       }
     }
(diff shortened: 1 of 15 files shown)


okeefm:
@rbclark May also be worth building some Cucumber tests for the button functionality, if you have time.

app/models/product.rb
```diff
@@ -140,6 +142,18 @@ def add_filtering_tests
...
+  # When the 2014 certification edition is enabled, duplicate_records is always disabled.
+  # Since the default for duplicate_records is true, we override it here. We also make sure
+  # that c4_test is not somehow set to true.
+  def enforce_cert_edition_settings
+    if cert_edition.eql? '2014'
+      self.duplicate_records = false
+      self.c4_test = false
+    end
+
+    true
```
okeefm: What is the purpose of this `true` return statement here?

rbclark: before_* always have to return true, if that is not there then `self.c4_test = false` is that last statement which makes the return code `false`.

app/assets/javascripts/measure_selection.js
```diff
@@ -269,6 +268,7 @@ ready_run_once = function() {
       }
       else if (edition == '2015') {
         setCheckboxDisabled('#product_duplicate_records', false);
+        $('#product_duplicate_records').prop("checked", true);
```
okeefm: We also need to add the logic so that this Duplicate Records box is only checked if C2 is checked, and is unchecked if C2 is unchecked, as discussed between myself, @rbclark and @dczulada, before pulling in.

rbclark:
@okeefm There should now be additional cucumber tests, let me know if you think they're sufficient.

PR #958, opened 2018-01-05
Maintain measure test table order on updates

ljades:
The Measure Test table would previously jump out of order when viewing it while it was building. An example would be that tests would jump to the order in which they would be built during building. This fix addresses that and ensures they maintain sorting by CMS ID. Additionally, it fixes another issue where the Datatable would not be reinitialized, thus the ordering options at the top would not be present until a refresh or building/testing is done.

Before:
Start of build -
image
Shortly after -
image

After:
The following set of tests maintains this order the whole time, along with the Datatables styling and sorting options -
image

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-133
- [x] Tests are included and test edge cases (N/A)

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


ljades:
A test is failing on Travis that passes locally and seems disconnected from my work. I'm on my phone and can't rebuild from here though, but I can do so later to double check.

okeefm:
After a rebase, I'm getting this error:

screen shot 2018-02-16 at 10 33 18 am

app/assets/javascripts/application.js
```diff
@@ -29,7 +29,7 @@
 //= require assets_framework/breadcrumb
 //= require jquery_nested_form
 //= require jasny-bootstrap.min
-//= require local_time
+//= require local-time
```
okeefm: This change needs to get reverted, after the rebase.

PR #1038, opened 2018-09-12
Minimal patch

dczulada:
Enables the api measure evaluator
Expected Results Calculator made more generic (instead of product-test centric)
Code System Name for Race and Ethnicity set in config file (CYPRESS-1461)
CQM-Models updated to latest QDM 5.3 release (0.8.4).  Migration script to add QDM::Ids to existing patients
Patient Filter was not working properly, always returned false
Fixed a schematron assertion that appears to break with nokogiri on a mac
Update to point to an updated cam-parsers branch for QRDA1 export.

**Submitter:**
- [x] Internal ticket for this PR:
- [ ] Tests are included and test edge cases

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


lib/cypress/cql_bundle_importer.rb
```diff
@@ -103,18 +103,19 @@ def self.unpack_and_store_qdm_patients(zip, type, bundle)
       puts "\rLoading: Patients Complete          "
     end
 
+    # TODO: This will need to be updated for 2018.0.2 bundles that store relatedTo as an QDM::ID
```
okeefm: Does this `TODO` need to get resolved before the patch gets merged?

dczulada: Originally I was thinking that we would export the bundle using the new 0.8.4 model.  I now believe it is far less complex to export the bundle in the same 0.8.2 format and deal with the conversion on import.  That way both bundle will be imported the same way.  Fewer moving pieces.  I'll remove the TODO.

lib/cypress/api_measure_evaluator.rb
```diff
@@ -86,12 +86,12 @@ def run_vendor_tests(vendor_link, measures, product_name, skip_c1_test, bundle_i
...
         calcuate_cat_3(patient_links[0].split('/')[2], bundle_id)
         upload_test_execution(extract_test_execution_link(patient_links[1], 'C1'), patient_links[0].split('/')[2], true) unless skip_c1_test
-        # upload_test_execution(extract_test_execution_link(patient_links[1], 'C2'), patient_links[0].split('/')[2], false, skip_c1_test)
+        upload_test_execution(extract_test_execution_link(patient_links[1], 'C2'), patient_links[0].split('/')[2], false, skip_c1_test)
       end
-      # sleep(4)
-      # download_filter_data
-      # calculate_filtered_cat3(bundle_id)
-      # upload_c4_test_executions
+      sleep(4)
```
rbclark: Would it be possible to verify the data is in the expected state here instead of sleeping?

test/models/product_test_test.rb
```diff
@@ -126,7 +126,7 @@ def compare_records(patient1, patient2)
       de2 = patient2.dataElements.fetch(x)
       de1.attributes.each do |k, v|
         assert_nil de2.attributes[k], 'random repeatability error: dataElements different, non-nil match' if v.nil?
-        assert_equal v, de2.attributes[k], 'random repeatability error: dataElements different' unless k == '_id'
+        assert_equal v, de2.attributes[k], 'random repeatability error: dataElements different' unless %w[_id id].include?(k)
```
rbclark: Could you document here why we're getting both `id` and `_id` attributes here for clarity?

PR #1084, opened 2019-01-02
Address Cypress-369 where code and negation reason could be in differ…

dczulada:
…ent entries

In record sample test: Code is checked for by looking through the whole file instead of just expected template.

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-369

**Reviewer 1:**

Name: @mayerm94 

**Reviewer 2:**

Name: Lauren DiCristofaro


lib/validators/attribute_extractor.rb
```diff
@@ -16,20 +16,19 @@ module AttributeExtractor
 
     # does a node need to be checked for an atttibute value
     # sc = source_criteria, cc = checked_criteria
```
mayerm94: sc is no longer used

test/fixtures/qrda/checklist/multiple_entries.xml
```diff
@@ -0,0 +1,64 @@
+<section xmlns="urn:hl7-org:v3"
```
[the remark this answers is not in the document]
dczulada: They are QRDA snippits from selected Cypress patient who have the data elements/attributes of interest for the C1 Record Sample tests.  There is no reason why moving forward these could not be created/exported from bonnie.

lib/validators/checklist_result_extractor.rb
```diff
@@ -4,102 +4,96 @@ module ChecklistResultExtractor
...
       end
       if passing
-        checked_criteria.passed_qrda = true
-        checked_criteria.save
+        @checked_criteria.passed_qrda = true
+        @checked_criteria.save
       end
     end
 
-    def passing_dc?(codenodes, source_criteria, checked_criteria)
```
lmd59: This looks like a paradigm change to expecting the internal state to track the current source_criteria, checked_criteria. I think that is okay because this and below methods are used only by find_dc_node, which establishes the state. But if we have the expectation that all of these methods will only ever be called internally, we should mark them as private.

PR #1101, opened 2019-02-06
Restructure Patient model for easier integration of Bring your own Patient

rbclark:
This adds 3 subclasses of Patient, BundlePatient, VendorPatient, and ProductTestPatient

These 3 subclasses allow us to take advantage of Rails Single Table Inheritance, providing us with a :_type field on the Patient model. This type field allows us to properly use has_many relations in the app again which was previously not possible on models such as the bundle model.

Additionally, this has restructured the bundle import to use insert_many instead of insert_one which provides a huge performance speedup during bundle import.

**Submitter:**

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:

### DIFF
--- a/lib/cypress/population_clone_job.rb
+++ b/lib/cypress/population_clone_job.rb
@@ -81,7 +81,6 @@ module Cypress
         original_patient_id: patient.id,
         correlation_id: options['test_id']
       }
-      # TODO: R2P: use patient model
       unnumerify cloned_patient if patient.givenNames.map { |n| n =~ /\d/ }.any? || patient.familyName =~ /\d/
       cloned_patient.medical_record_number = next_medical_record_number unless options['disable_randomization']
       DemographicsRandomizer.randomize(cloned_patient, prng, allow_dups) if options['randomize_demographics']
--- a/lib/cypress/population_clone_job.rb
+++ b/lib/cypress/population_clone_job.rb
@@ -76,12 +76,14 @@ module Cypress
       # If we need to clone a patient for any other reason then we will need to paramaterize
       # the type coming into this class.
       cloned_patient = ProductTestPatient.new(patient.attributes.except(:_id, :_type))
+      cloned_patient.attributes = {
+        original_medical_record_number: patient.medical_record_number,
+        original_patient_id: patient.id,
+        correlation_id: options['test_id']
+      }
       # TODO: R2P: use patient model
       unnumerify cloned_patient if patient.givenNames.map { |n| n =~ /\d/ }.any? || patient.familyName =~ /\d/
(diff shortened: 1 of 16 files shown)


lib/cypress/population_clone_job.rb
```diff
@@ -72,13 +72,18 @@ def randomize_ids(patients, prng)
...
+      # This operates on the assumption that we are always cloning a patient for a product test.
+      # If we need to clone a patient for any other reason then we will need to paramaterize
+      # the type coming into this class.
+      cloned_patient = ProductTestPatient.new(patient.attributes.except(:_id, :_type))
+      cloned_patient.attributes = {
+        original_medical_record_number: patient.medical_record_number,
+        original_patient_id: patient.id,
+        correlation_id: options['test_id']
+      }
       # TODO: R2P: use patient model
```
okeefm: Can this `TODO` be removed now?

lib/validators/calculating_augmented_records.rb
```diff
@@ -20,7 +19,7 @@ def parse_and_save_record(record)
     # create a temporary saved record copy to do calculations on, then delete it
```
okeefm: Should change this comment, now that the record is no longer saved.

lib/ext/patient.rb
```diff
@@ -135,4 +140,15 @@ def randomize_demographics(patient, changed, random: Random.new)
     # HDS helpers
     #
   end
+
+  class BundlePatient < Patient; end
+
+  class VendorPatient < Patient; end
```
okeefm: I assume `VendorPatient` is associated with the BYOPatients? If so, I'm ok with waiting on adding tests for them.

rbclark: VendorPatient is associated with BYOPatients, as soon as we have the actual functionality we can add tests but at this time a VendorPatient is really just nothing more than a `_type: CQM::VendorPatient` field in the DB.

okeefm: That's what I figured. Sounds good!

lib/validators/calculating_augmented_records.rb
rbclark: Looking back at this we should maybe make a few more considerations here, `parse_record` passes over to CqmExecutionCalc which saves IndividualResults to the DB and also returns the individual results. Do we always need to save IndividualResults to the DB/do we ever need IndividualResults saved to the DB?

lib/cypress/population_clone_job.rb
rbclark: I believe so, I've gone ahead and removed it.

PR #1105, opened 2019-02-15
Docker for Cypress v5

okeefm:
This PR fixes the Cypress Docker and Docker Compose infrastructures to work with Cypress v5, as currently available in `master`. It updates:

* The Ruby base image used to build Cypress
* The measure calculation engine used by Cypress (to `cqm-execution-service`, as built on https://hub.docker.com/r/tacoma/cqm-execution-service
* The way the CQM Execution Service URL gets built, to allow for it to be on a host other than `localhost` (useful for more than Docker)
* A typo (I think?) that caused a separate temp folder (`temp/bundles`) to be built, rather than using the existing `tmp` folder
* An environment variable (`DISABLE_DB`) to be used to kill DB access during the precompile step, to allow for precompilation of assets in the Docker image
* One update to `require` the `BootstrapBreadcrumbsBuilder`, which wasn't previously included.

**Submitter:**
- [x] Internal ticket for this PR: N/A
- [x] Internal ticket links to this PR N/A
- [x] Tests are included and test edge cases N/A
- [x] Tests have been run locally and pass N/A

**Reviewer 1:**

Name: Robert Clark

**Reviewer 2:**

Name:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,9 @@
-# WARNING: This Dockerfile, and the associated docker-compose.yml,
-# are only functional for Cypress v3.x. Docker installation 
-# methods are not yet supported for Cypress v4, as of 2018-07-10
-
 FROM phusion/passenger-ruby23:latest
 
 RUN apt-get update \
     && apt-get upgrade -y -o Dpkg::Options::="--force-confold" \
     && apt-get install -y nodejs tzdata \
     && rm -rf /var/lib/apt/lists/*
-
-# Install ruby 2.3.4 in order to match what we are developing against.
-RUN bash -lc 'rvm install ruby-2.3.4'
-RUN bash -lc 'rvm --default use ruby-2.3.4'
-RUN gem install bundler
 
 ENV RAILS_ENV production
 ENV RAILS_SERVE_STATIC_FILES 1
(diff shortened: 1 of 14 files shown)


Dockerfile
```diff
@@ -1,19 +1,11 @@
-# WARNING: This Dockerfile, and the associated docker-compose.yml,
-# are only functional for Cypress v3.x. Docker installation 
-# methods are not yet supported for Cypress v4, as of 2018-07-10
-
-FROM phusion/passenger-ruby23:latest
+# Pinned to the latest ruby 2.6.0 version of the Passenger base Docker image
+FROM phusion/passenger-ruby26:1.0.1
```
rbclark: Do you mind expanding the comment here to say when we can actually bump to a newer version, is there a specific issue blocking us from using latest? Is this just due to the fact that we're using a slightly out-of-date version with Cypress?

config/cypress.yml
```diff
@@ -2,7 +2,7 @@
 # upgrades failing to complete successfully.
 
 # These settings are kept in the constants file with the assumption that the end user should not be overriding them.
-bundle_file_path: 'temp/bundles'
+bundle_file_path: 'tmp/bundles'
```
rbclark: Nice, always wondered why there was a temp and tmp directory but I never dug into why. As long as the bundle files are never used after import this should be fine, we do run a `tmp:cache:clear` on upgrades, so that is the only potential problem I can think of.

config/initializers/mongoid.rb
```diff
@@ -8,4 +8,6 @@ def as_json(*args)
       to_s.as_json
     end
   end
-end
\ No newline at end of file
+end
+
+Mongoid::Tasks::Database.create_indexes if Rails.env.production? && !ENV['DISABLE_DB']
```
rbclark: It may not hurt to create indexes here on development and production. I definitely would appreciate the indexes creating automatically during development for me.

Dockerfile
okeefm: The only restriction here is that we need to use a version of `passenger-ruby26` that matches the `.ruby-version` version of Ruby (at the moment, 2.6.0). Phusion only tracks minor version bumps in package names, so we can't bump to `1.0.2` (or `latest`) until we upgrade Cypress to 2.6.1.

config/cypress.yml
okeefm: AFAIK the bundle files are never used after import, so we should be safe here.

config/initializers/mongoid.rb
okeefm: That would be of use outside of the Docker ecosystem, correct? Seems like a worthwhile enhancement.

Dockerfile
rbclark: Sounds good, once this is merged we should make a new PR to bump ruby to 2.6.1 across the board.

config/initializers/mongoid.rb
rbclark: Yeah that would enhance it for all developers as well. Of note, this change might also affect the Pkgr builds, we currently run this script but it seems it can now be removed: https://github.com/projectcypress/cypress/blob/master/contrib/scripts/after_install.sh

okeefm: The only thing that script does that the Mongoid initializer doesn't is the `db:migrate` task. Does that still need to get run for Pkgr builds?

rbclark: I think it sounds like we came to the conclusion that we don't need the db:migrate at this time, so we should remove it from both, correct?

okeefm: Sounds good to me. I removed it from `after_install`, and didn't add it here.

rbclark: I built a full VM using this pull request and the automatic settings detection worked correctly, so this has not broken any of the production environment related stuff.

rbclark:
@luis1van You are correct that it is not rendering correctly but I think you got the wrong line, when you import a bundle, create a vendor and create a patient and then click on "Download All Patients", the button is currently not working in production. It is due to the asset requested on https://github.com/projectcypress/cypress/blob/e31f0e8f48a7aeea447ed729031bdaf68f43bb2c/app/views/layouts/report.html.erb#L14. This causes an `undefined method find_asset for nil:NilClass`.

okeefm:
@luis1van Good find on that bug. I pushed up a fix to that (and a fix to the docker-compose.yml that makes the cypress volume persist)

okeefm:
I believe the cqm-execution-service should now rebuild itself automagically, as I've added the Docker Hub webhook to that repo. See https://hub.docker.com/r/tacoma/cqm-execution-service.

rbclark:
Verified that the problem Luis found has in fact been corrected.

PR #1114, opened 2019-03-05
Byo patient upload patients

rbclark:

**Submitter:**
- [ ] This pull request describes why these changes were made.
- [ ] Code diff has been done and been reviewed
- [ ] Tests are included and test edge cases
- [ ] Tests have been run locally and pass

**Reviewer 1:**

Name: @mayerm94 

**Reviewer 2:**

Name:


app/views/records/_records_list.html.erb
```diff
@@ -4,6 +4,10 @@
   <% result_values = get_result_values(patients, [measure], measure.population_ids.keys(), 'patient_id') %>
 <% end %>
 <h1><%= measure ? measure.display_name : 'All' %> Patients</h1>
+<% if @vendor %>
+  <% patients= @vendor.patients %>
+  <% bundle= Bundle.all.to_a.first #temp for test%>
```
mayerm94: Why cast to an array and get the first rather than just get the first out of the Bundle collection? Also is the comment on this still applicable, and if it is... does that mean that this line is no longer needed?

lmd59: Good catch! Yes- this line should be removed. PR now updated as such.

test/factories/bundles.rb
```diff
@@ -4,6 +4,8 @@
     sequence(:version) { |i| "#{2017 + i}.0.0" }
     sequence(:title) { |i| "Bundle Name #{i}" }
 
+    # static_bundle includes random measures that will not execute in the cqm-execution-service
```
mayerm94: Are these random measures non-executable intentionally? Or do we just not know why they don't work?

PR #1123, opened 2019-03-20
Multi measure models

dczulada:
This will still need more tests to be added.  However, since it will require more measure fixtures and test it probably makes more sense to wait until after the measure model updates are pulled in to do those.

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-343

**Reviewer 1:**

Name: @okeefm 

**Reviewer 2:**

Name:


app/views/test_executions/_test_info.html.erb
```diff
@@ -6,14 +6,20 @@
 
 <div class = 'col-sm-4'>
   <h1><%= info_title_for_product_test(task.product_test) %></h1>
-  <% unless task.product_test.is_a? ChecklistTest %>
+  <% if (task.product_test.is_a? MeasureTest) || (task.product_test.is_a? FilteringTest) %>
     <strong>Measure: </strong><span><%= task.product_test_name %></span><br/>
-  <% end %>
-  <strong>HQMF ID: </strong><span><%= task.measure_ids.join(',') %></span><br/>
```
lmd59: It appears this line has been removed for the case where the test is a checklist test. Did we intend on removing the HQMF id from the "Record Sample Test Information" page?

dczulada: That was unintentional.  Thou...I'm not sure how useful having the HQMF id on the Record Sample Test Execution. (below is a screenshot from the demo server with the HQMF ids)

image

app/models/multi_measure_cat3_task.rb
```diff
@@ -0,0 +1,22 @@
...
+  def validators
+    @validators = [::Validators::MeasurePeriodValidator.new,
+                   ::Validators::QrdaCat3Validator.new(product_test.expected_results, true, true, false, product_test.bundle),
+                   ::Validators::CMSQRDA3SchematronValidator.new(product_test.bundle.version, false),
+                   ::Validators::ExpectedResultsValidator.new(product_test.expected_results)]
+    @validators
+  end
+
+  def execute(file, user)
+    te = test_executions.create!(expected_results: expected_results, artifact: Artifact.new(file: file), user_id: user)
```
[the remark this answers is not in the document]
okeefm: I'm not sure we're changing anything on the user here that would require saving it; the way the `has_many`/`belongs_to` relation works in Rails is that the relation ID is kept on the element that belongs_to, in this case the Test Execution.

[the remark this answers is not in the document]
dczulada: The user is associated with the test execution at the initialize step. This is slightly different than some of the other tasks, causing confusion

app/views/products/_cvu_plus_product.html.erb
```diff
@@ -0,0 +1,69 @@
...
+
+<div class="product-test-tabs">
+
+<div class="product-test-tabs">
+  <ul>
+    <% each_tab(@product) do |test_type, title, _, html_id| %>
+      <li><a href = '<%= "##{html_id}" %>'><%= title %></a></li>
+    <% end %>
+  </ul>
+  <% each_tab(@product) do |test_type, title, description, html_id| %>
```
okeefm: Isn't there only one tab in this scenario? Or are we anticipating adding additional test types as CVU+ matures?

PR #1130, opened 2019-04-08
Cqm models master

dczulada:
This PR updates Cypress to use the master branch of cqm-models.

Three aspects of Cypress were impacted. Changes to these three aspects are made in distinct commits.

1) Updates to the provider model.  A CQM:Provider is not embedded in a CQM::Patient, before, this was a provider_performance which referenced a provider.  This provider performance was never really used for anything on its own.
2) CQM::IndividualResult is now QDM::IndividualResult.  and cqm-patient alias was created so it could reference the CQM::Patient, not the QDM::Patient
3) A measure's Source_Data_Criteria is now an array for QDM::DataCriteria, instead of an HDS hash.  Source Data Criteria is used heavily is the C1 Record Sample tests, C4 Filter tests (with the 'Problem' filter) and Scoop and Filter.

**Submitter:**
- [x] Internal ticket for this PR: https://jira.mitre.org/browse/CYPRESS-467

**Reviewer 1:**

Name: @okeefm 

**Reviewer 2:**

Name:


lib/ext/individual_result.rb
```diff
@@ -1,10 +1,16 @@
...
+module QDM
   class IndividualResult
     store_in collection: 'individual_results'
 
     field :population_set_key, type: String
     field :correlation_id, type: String
+
+    belongs_to :cqm_patient, class_name: 'CQM::Patient', inverse_of: :calculation_results
+
+    def patient
```
okeefm: Might be worth a comment describing the function of this function

test/unit/lib/record_filter_test.rb
```diff
@@ -234,7 +234,8 @@ def test_provider_filter
     prov = Provider.default_provider
 
     patients.each do |patient|
-      patient.provider_performances = [{ provider_id: prov.id }]
+      patient.providers = nil
```
okeefm: Does this work? Wouldn't it have to be `patient.providers = []`, so the provider could get pushed onto it?

dczulada: both approaches seem to work equivalently

has_and_belongs_to_many :providers, class_name: 'CQM::Provider'

my guess is mongoid/activerecord magic with "has_and_belongs_to_many"

lib/validators/checklist_criteria_validator.rb
```diff
@@ -31,26 +31,26 @@ def validate(file, _options = {})
...
-        template = oids.map { |oid| QRDA::Util::QRDATemplateHelper.definition_for_template_id(oid, @qrda_version) }.compact.first['cda_template_id']
+        template = QRDA::Util::QRDATemplateHelper.definition_for_template_id(oid, @qrda_version)['cda_template_id']
         # find all nodes that fulfill the data criteria, this is defined in checklist result extractor
         find_dc_node(template, checked_criteria, sdc)
       end
     end
 
-    def validate_demographics(hqmf_oids, checked_criteria)
+    def validate_demographics(sdc, checked_criteria)
       xpath_map = {
```
lmd59: Do we want to put xpath codes in a shared (i.e. library, global constants, or config) area so that they can be consistently used across this project (or multiple projects as necessary)

PR #1195, opened 2019-07-08
Find an appropriate data element description with scoop and filter

dczulada:
Adds 'text' to xml 
image

and adds  to html

image

**Submitter:**

**Reviewer 1:**

Name:


lmd59:
Would potentially recommend putting the text code name *inside* the code section of the html  patient view (instead of above), but looks good/very helpful!

PR #2048, opened 2026-05-15
Product test page speed up and other performance improvements

elsaperelli:
## Summary
This PR makes some application performance improvements and unit test updates. The main goal of this PR is to speed up product/vendor pages by caching the product test task status.

## Code Changes
- The most recent task status is now stored directly on the `ProductTest` (`most_recent_product_test_status` and `most_recent_product_test_status_updated_at`) and this is what is now used in product status rendering
- Rake task added to backfill cached task statuses for existing product tests
- Make sure product test status is refreshed with updates to `Task`
 and `TestExecution`
- Vendor page non-favorite product loading with Mongo instead of loading and rejecting favorites in Ruby 
- Found issue in `ProductTestSetupJob`- filtering tests now skip the IPP retry behavior that was causing issues when that behavior is meant for measure tests

**Submitter:**
- [X] This pull request describes why these changes were made.
- [X] Code diff has been done and been reviewed
- [X] Tests are included and test edge cases
- [X] Tests have been run locally and pass

**Reviewer 1:**

Name:

**Reviewer 2:**

Name:


app/models/task.rb
```diff
@@ -42,19 +46,29 @@ class Task
   end
 
   def status
```
dczulada, line 48: For backward compatibility this does work, however it will still leave older Product Tests without the **most_recent_product_test_status** hash.

Would it make sense to do something like this?  This will add the most_recent_product_test_status to the tests that were run prior to the upgrade.


```
  def status
    current_status = current_cached_status
    return current_status unless current_status.nil?
    
    refresh_product_test_status
    product_test&.reload
    current_cached_status
  end

  def current_cached_status
    status = product_test&.most_recent_task_status(_type)
    status if status && _type != 'Task'
  end
```

dczulada:
The "Certification Status Table" doesn't seem to be backward compatible for test created prior to the upgrade.  The C1/C2 statuses seem to show up well, but the C3 ones don't.

Screenshot 2026-05-22 at 3 48 08 PM

Locally I added this code at the top of _certification_status_table.html.erb to run the rebuild method if the cached status values don't exist for the product test.

<% if product.product_tests.first.tasks.first.current_cached_status.nil? %>
    <% product.product_tests.each(&:rebuild_most_recent_product_test_status!) %>
  <% end %>

dczulada (changes requested):

dczulada (approved):

